PR #29, opened 2015-05-24
trimNodeRange in conditionals

ef4:
If you have a function that ends with a conditional, the close-of-function `})` can get written before the close-of-conditional `}`, due to the fact that functions trim trailing whitespace and conditionals do not.

This adds trimming to conditionals. It's possible that other block forms suffer the same bug and will also need to start trimming in case they appear at the end of a function.


eventualbuddha:
Thanks! Released in v1.16.0.

PR #31, opened 2015-05-24
Fix return of interpolated string

ef4:
Because of the way we are doing string-replacement and not AST->AST transformation, we are quite sensitive to the order the transforms run.

In this case, if you rewrite the `->` and the `"#` first, the patcher can't find any original characters to target when it tries to insert the `return`.


eventualbuddha:
Thanks!

eventualbuddha:
Released in v1.16.0.

PR #33, opened 2015-06-19
Update traverse.js

dolzenko:


eventualbuddha:
Thanks! Would you mind adding a test for both of these that show that they're passed through?

eventualbuddha:
Nevermind, I added the tests in 70549cf. Released as part of v1.16.1.

PR #55, opened 2015-09-16
Use https in package.json for git repository

netei:
This makes it work when behind a proxy

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   "author": "Brian Donovan",
   "license": "MIT",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


eventualbuddha:
Looks like node 0.10 doesn't like this change. Eith node 4.0 out I think I'll change the tests to run on 0.12 and 4.0. Then this should be mergeable.

netei:
It probably is because of the default version of npm on node v0.10.

Switching to 0.12 or updating npm (with `npm udpate npm -g`) should do the trick indeed

eventualbuddha:
Thanks @netei!

PR #70, opened 2015-11-26
Fix in operator to work in compound expressions

jeffchan:
First time contributing, so let me know if this is not the way to fix it :)


eventualbuddha:
Thanks a lot! This is, unfortunately, the right way to fix it.

eventualbuddha:
Released in v1.40.8.

PR #71, opened 2015-11-27
Ensure that hashes in regexp are not treated as comments

jeffchan:
Not sure if this is the best way to fix. Would love to get feedback


eventualbuddha:
Looks great! Thanks.

eventualbuddha:
Released in v1.40.8.

Issue #72, opened 2015-11-30
Error: BUG! Could not fix range for SubtractOp because it has no raw value


jcalfee:
I'm not sure where the error is, but this is how to re-create it:
```bash
curl https://raw.githubusercontent.com/cryptonomex/graphene-ui/master/dl/src/ecc/signature.coffee|decaffeinate
```
Produces the following error:
```bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  3601  100  3601    0     0   8383      0 --:--:-- --:--:-- --:--:--  8393
/home/jcalfee/opt/iojs-v2.0.1-linux-x64/lib/node_modules/decaffeinate/lib/utils/parse.js:110
      throw new Error('BUG! Could not fix range for ' + node.type + ' because 
            ^
Error: BUG! Could not fix range for SubtractOp because it has no raw value
    at fixRange (/home/jcalfee/opt/iojs-v2.0.1-linux-x64/lib/node_modules/decaffeinate/lib/utils/parse.js:110:13)
    at /home/jcalfee/opt/iojs-v2.0.1-linux-x64/lib/node_modules/decaffeinate/lib/utils/parse.js:43:5
    at traverse (/home/jcalfee/opt/iojs-v2.0.1-linux-x64/lib/node_modules/decaffeinate/lib/utils/traverse.js:35:23)
    at /home/jcalfee/opt/iojs-v2.0.1-linux-x64/lib/node_modules/decaffeinate/lib/utils/traverse.js:30:9
    at Array.forEach (native)
    at descend (/home/jcalfee/opt/iojs-v2.0.1-linux-x64/lib/node_modules/decaffeinate/lib/utils/traverse.js:21:32)
    at traverse (/home/jcalfee/opt/iojs-v2.0.1-linux-x64/lib/node_modules/decaffeinate/lib/utils/traverse.js:38:5)
    at /home/jcalfee/opt/iojs-v2.0.1-linux-x64/lib/node_modules/decaffeinate/lib/utils/traverse.js:26:11
    at Array.forEach (native)
    at /home/jcalfee/opt/iojs-v2.0.1-linux-x64/lib/node_modules/decaffeinate/lib/utils/traverse.js:24:15
```

PR #75, opened 2015-12-02
Fix indent detection

eventualbuddha:
This still doesn't fix the sample file in #72 as there are further bugs with comment handling. More PRs forthcoming.

cc @jeffchan @juliankrispel @nickdima (let me know if you prefer not to be cc'd on these)

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,8 +30,9 @@
   "license": "MIT",
   "dependencies": {
     "coffee-script-redux": "https://github.com/michaelficarra/CoffeeScriptRedux.git",
+    "detect-indent": "^4.0.0",
     "magic-string": "^0.6.5",
-    "repeat-string": "^1.5.2"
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -30,8 +30,9 @@
   "license": "MIT",
   "dependencies": {
     "coffee-script-redux": "https://github.com/michaelficarra/CoffeeScriptRedux.git",
+    "detect-indent": "^4.0.0",
     "magic-string": "^0.6.5",
-    "repeat-string": "^1.5.2"
+    "repeating": "^2.0.0"
```
jeffchan: Just curious why the switch?

eventualbuddha: `repeating` is a dependency of `detect-indent`, so we're already bundling it. Might as well consolidate!

PR #76, opened 2015-12-03
Unify block & doc comments.

eventualbuddha:
This changes the behavior of block comments such that leading hash marks no longer mark a comment as being a documentation comment. Instead, all block comments are treated equally and users wishing to make a block comment a doc comment simply add an asterisk after the initial `###`.


eventualbuddha:
Released in v1.40.15.

PR #77, opened 2015-12-03
Transform static class methods.

eventualbuddha:
Addresses #72.


eventualbuddha:
Released in v1.40.15.

Issue #86, opened 2015-12-08
Does not work with heregexes and comments


netei:
Input : 

```
regex= ///
[^a-z]+  #foo
[a-z]+
///
```

Output: 

```
~/.nvm/versions/node/v0.12.7/lib/node_modules/decaffeinate/dist/decaffeinate.cjs.js:813
    throw new Error('BUG! Could not fix range for ' + node.type + ' at line '
          ^
Error: BUG! Could not fix range for Program at line 1, column 1
    at fixRange (~/.nvm/versions/node/v0.12.7/lib/node_modules/decaffeinate/dist/decaffeinate.cjs.js:813:11)
    at ~/.nvm/versions/node/v0.12.7/lib/node_modules/decaffeinate/dist/decaffeinate.cjs.js:723:5
    at traverse (~/.nvm/versions/node/v0.12.7/lib/node_modules/decaffeinate/dist/decaffeinate.cjs.js:473:23)
    at parse (~/.nvm/versions/node/v0.12.7/lib/node_modules/decaffeinate/dist/decaffeinate.cjs.js:721:3)
    at convert (~/.nvm/versions/node/v0.12.7/lib/node_modules/decaffeinate/dist/decaffeinate.cjs.js:4861:13)
    at ReadStream.<anonymous> (~/.nvm/versions/node/v0.12.7/lib/node_modules/decaffeinate/dist/decaffeinate.cjs.js:4816:16)
    at ReadStream.emit (events.js:129:20)
    at _stream_readable.js:908:16
    at process._tickCallback (node.js:355:11)
```

netei:
Doesn't work with the following herexegp either : 

```
"foo".replace ///
foo
///g, "bar"
```

eventualbuddha:
This appears to be two different issues. The heregex with the comment fails due to a parsing bug with CoffeeScriptRedux. Namely, it seems to strip the comment and treat it as if that part of the string were never there:

```js
{ type: 'Program',
  line: 1,
  column: 1,
  raw: 'regex= ///\n[^a-z]+  \n[a-z]+\n///',
  range: [ 0, 31 ],
  body: 
   { type: 'Block',
     line: 1,
     column: 1,
     raw: 'regex= ///\n[^a-z]+  \n[a-z]+\n///',
     range: [ 0, 31 ],
     statements: [ [Object] ] } }
```

That makes the `range` values for the nodes incorrect, which leads to the error you encountered.

[...]

Issue #65, opened 2015-10-03
Switch to official CoffeeScript parser


eventualbuddha:
Due to the [various bugs](https://github.com/eventualbuddha/decaffeinate/issues?q=is%3Aopen+is%3Aissue+label%3Aredux-bug) with parsing valid CoffeeScript I've decided to switch parsers from CoffeeScriptRedux ([coffee-script-redux](https://www.npmjs.com/package/coffee-script-redux)) to the official CoffeeScript parser ([coffee-script](https://www.npmjs.com/package/coffee-script)).

decaffeinate only knows how to operate on the AST generated by CoffeeScriptRedux, which means I either have to do significant rewrites of decaffeinate or turn the AST generated by CoffeeScript into one that matches what CoffeeScriptRedux would generate. I started working on a proof-of-concept for the latter called [decaffeinate-parser](https://github.com/eventualbuddha/decaffeinate-parser), and it is now capable of mapping 99% of the AST generated for the CoffeeScript project source code itself (including tests).

To make the transition complete, I'll need to:

- [x] Replace coffee-script-redux with coffee-script
- [ ] Fix any regressions introduced by that change
- [ ] Remove the warning from the README about CSR and syntax incompatibility
- [ ] Update the "Status" of the project to something slightly more confident-sounding than "Incomplete"

blimmer:
Big fan of this!

nickdima:
A bit ironic investing time in something like this instead of making CoffeeScript (and its ecosystem) better. 
No pun intended just a thought.

blimmer:
I rely on this tool because my webapps use EmberJS and the community has rallied behind Babel transpiling. This tool is extremely valuable.

nickdima:
Yes, I can see that. A lot of us moving from CoffeeScript to ES6 lately, not mainly for ES6 itself but more for the tools and the community.

sandstrom:
Just wanted to say that decaffeinate is a great tool, thanks! :boat:

nickdima:
@eventualbuddha is `decaffeinate-parser` ready to be used in `decaffeinate`?
Would like to help on the project if possible.

eventualbuddha:
@blimmer & @sandstrom Thanks for the kind words!

[...]

PR #88, opened 2015-12-08
Fix trimming heregexes.

eventualbuddha:
Addresses #86.

This is quickly becoming a poor man’s CoffeeScript lexer. Hopefully this can be obviated by #65.

cc @jeffchan @netei


jeffchan:
:+1: Agreed

PR #127, opened 2016-02-25
Add UnaryExistsOpPatcher

### LINKED ISSUE #126: New patcher for UnaryExistsOp
First of, thanks so much for writing all this, it's super useful!

I'm using the patchers branch with the decaf-parser and alot of the code I'm currently converting has lots of
```
if data.promocode?
  stuff()

if data.blah?
  something_else()
```

to convert this it looks like a patcher needs to be written for  `UnaryExistsOp` and I'm happy to write it

I've had a look through the existing patchers but something isn't quite clicking.. are there any tips or pointers you could give, or at least suggest a good patcher to start from.

Thanks!

jeffchan:
Closes #126

First stab at this, many thanks to https://github.com/decaffeinate/decaffeinate/blob/master/CONTRIBUTING.md (very helpful)

Any feedback?

### DIFF
--- a/src/patchers/UnaryExistsOpPatcher.js
+++ b/src/patchers/UnaryExistsOpPatcher.js
@@ -1,0 +1,60 @@
+import UnaryOpPatcher from './UnaryOpPatcher.js';
+import type { Node, ParseContext, Editor, NodePatcher } from './types.js';
+
+export default class UnaryExistsOpPatcher extends UnaryOpPatcher {
+  constructor(node: Node, context: ParseContext, editor: Editor, expression: NodePatcher) {
+    super(node, context, editor, expression);
+    this.negated = false;
+  }
+
+  patchAsExpression() {
+    let needsParens = !this.isSurroundedByParentheses();
+    if (needsParens) {
+      // `a or a?` → `a or (a?`
+      //       ^           ^
+      this.insertBefore('(');
+    }
+    this.patchAsStatement();
+    if (needsParens) {
+      // `a or (a?` → `a or (a?)`
+      //         ^             ^
+      this.insertAfter(')');
+    }
+  }
+
+  patchAsStatement() {
+    let { node, negated } = this;
+    let nodeExpression = node.expression;
+
+    if (nodeExpression && nodeExpression.type === 'Identifier') {
+      if (negated) {
+        // `a?` → `typeof a === 'undefined' && a === null`
+        //  ^^     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+        this.overwrite(this.start, this.end, `typeof ${nodeExpression.raw} === 'undefined' || ${nodeExpression.raw} === null`);
+      } else {
+        // `a?` → `typeof a !== 'undefined' && a !== null`
+        //  ^^     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+        this.overwrite(this.start, this.end, `typeof ${nodeExpression.raw} !== 'undefined' && ${nodeExpression.raw} !== null`);
+      }
+
+    } else {
+      if (negated) {
+        // `a.b?` → `a.b == null`
+        //     ^        ^^^^^^^^
+        this.overwrite(this.expression.after, this.end, ' == null');
+      } else {
+        // `a.b?` → `a.b != null`
+        //     ^        ^^^^^^^^
+        this.overwrite(this.expression.after, this.end, ' != null');
+      }
+    }
+  }
+
+  /**
+   * Flips negated flag but doesn't edit anything immediately so that we can
+   * use the correct operator in `patch`.
+   */
+  negate() {
+    this.negated = !this.negated;
+  }
+}
--- a/test/unary_operator_test.js
+++ b/test/unary_operator_test.js
@@ -40,17 +40,38 @@ describe('unary operators', () => {
     check(`typeof a`, `typeof a;`);
   });
 
-  it.skip('converts unary existential identifier checks to typeof + null check', () => {
+  it('converts unary existential identifier checks to typeof + null check', () => {
     check(`a?`, `typeof a !== 'undefined' && a !== null;`);
   });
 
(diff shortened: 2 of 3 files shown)


jeffchan:
Not sure why the Travis tests aren't running.

eventualbuddha:
I forgot to set up the Travis tests again after moving to the decaffeinate org. They should be working again.

src/patchers/UnaryExistsOpPatcher.js
```diff
@@ -0,0 +1,24 @@
...
+    // Only add parentheses if necessary
+    if (node.parentNode && node.parentNode.type !== 'Block') {
+        this.insertBefore('(');
+        this.insertAfter(')');
+    }
+
+    if (nodeExpression && nodeExpression.type === 'Identifier') {
+      this.overwrite(this.start, this.end, `typeof ${nodeExpression.raw} !== 'undefined' && ${nodeExpression.raw} !== null`);
+    } else {
+      this.overwrite(this.expression.after, this.end, ' != null');
```
eventualbuddha: I've started including comments here to explain more clearly what's going on. Here I'd put this:

```
// `a?` → `a != null`
//   ^      ^^^^^^^^
```

And same for the `Identifier` case.

src/patchers/UnaryExistsOpPatcher.js
```diff
@@ -0,0 +1,24 @@
+import UnaryOpPatcher from './UnaryOpPatcher.js';
+
+export default class UnaryExistsOpPatcher extends UnaryOpPatcher {
+  patchAsExpression() {
+    const node = this.node;
+    const nodeExpression = node.expression;
+
+    // Only add parentheses if necessary
+    if (node.parentNode && node.parentNode.type !== 'Block') {
```
eventualbuddha: This is pretty much trying to replicate the statement vs. expression thing. What I'd probably do is to put everything from `patchAsExpression` into `patchAsStatement` and then call the latter from the former like so:

```js
patchAsExpression() {
  this.insertBefore('(');
  this.patchAsStatement();
  this.insertAfter(')');
}
```

src/patchers/UnaryExistsOpPatcher.js
```diff
@@ -0,0 +1,24 @@
...
+        this.insertBefore('(');
+        this.insertAfter(')');
+    }
+
+    if (nodeExpression && nodeExpression.type === 'Identifier') {
```
eventualbuddha: For extra credit, you should handle negation a la [`EqualityPatcher`](https://github.com/decaffeinate/decaffeinate/blob/master/src/patchers/EqualityPatcher.js). You should add tests like this:

```js
check(`
  unless a?
    b
`, `
  if (typeof a === 'undefined' || a === null) {
    b;
  }
`);

check(`
  unless a.b?
    b
`, `
  if (a.b == null) {
    b;
  }
`);
```

jeffchan: Ok, will attempt this :smile:

jeffchan:
@eventualbuddha  -- getting stuck on this. I would've expected `isUnless` to be true. What am I missing? (Or perhaps I'm going down the wrong path)

```js
{ type: 'Conditional',
  line: 1,
  column: 1,
  range: [ 0, 13 ],
  isUnless: false,
  condition:
   { type: 'UnaryExistsOp',
     line: 1,
     column: 8,
     range: [ 7, 9 ],
     expression:
      { type: 'Identifier',
        line: 1,
        column: 8,
        range: [Object],
        data: 'a',
        raw: 'a',
        parentNode: [Circular],
        scope: Scope {} },
     raw: 'a?',
     parentNode: [Circular],
     scope: Scope {} },
  consequent:
   { type: 'Block',
     line: 2,
     column: 3,
     range: [ 12, 13 ],
     statements: [ [Object] ],
     raw: 'b',
     inline: false,
     parentNode: [Circular],
     scope: Scope {} },
  alternate: null,
  raw: 'unless a?\n  b',
  parentNode:
   { type: 'Block',
     line: 1,
     column: 1,
     range: [ 0, 13 ],
     statements: [ [Circular] ],
     raw: 'unless a?\n  b',
     parentNode:
      { type: 'Program',
        line: 1,
        column: 1,
        range: [Object],
        body: [Circular],
        raw: 'unless a?\n  b',
        scope: Scope {} },
     scope: Scope {} },
  scope: Scope {} }
```

eventualbuddha:
@jeffchan that does seem incorrect. Perhaps I did something wrong in decaffeinate-parser. I'll check. If decaffeinate-parser is doing its thing correctly you shouldn't have to look at the `Conditional` node at all, it'll call `negate()` on your patcher for you.

jeffchan:
Yeah, for what it's worth, I don't think `ConditionalPatcher` is visited in this case.

eventualbuddha:
@jeffchan should be fixed by 1cbae89eeec80dc90566adf3f9fe2ab628d151ea and https://github.com/decaffeinate/decaffeinate-parser/commit/3c0eb9d6e9071c69b9205be6c51870358e1d6b23.

test/unary_operator_test.js
```diff
@@ -40,17 +40,38 @@ describe('unary operators', () => {
...
+      if ((typeof a === 'undefined' || a === null) || !c) {
+        b;
+      }
+    `);
+
+    check(`
+      unless a.b?
+        b
+    `, `
+      if ((a.b == null)) {
```
jeffchan: The double parens is the last piece that I'm struggling with. Do I need to detect that the conditional has no surrounding parens, then I can assume that `ConditionalPatcher` will take care of it, so I should not insert the parens? What about compound expressions?

eventualbuddha: Ah. That's probably beyond the scope of this PR. Let's just test for the
double parens. What is the question about compound expressions?

jeffchan: Ah, nevermind. I was just rambling about how the solution I had in my head probably wouldn't work for compound expressions :stuck_out_tongue:

jeffchan:
@eventualbuddha Ready for another look!

eventualbuddha:
Thanks!

PR #132, opened 2016-02-26
Make logical compound assignment statements more natural.

### LINKED ISSUE #130: Use `if` statements for expanded compound assignment
For example, `a ||= b` currently becomes `a || (a = b);`, but a better translation when building a statement is `if (!a) { a = b; }`. The expression translation should be left alone.

/cc @jeffchan @sandfox

eventualbuddha:
Closes #130.

/cc @jeffchan @sandfox

### DIFF
--- a/src/patchers/LogicalOpCompoundAssignOpPatcher.js
+++ b/src/patchers/LogicalOpCompoundAssignOpPatcher.js
@@ -1,7 +1,7 @@
 import CompoundAssignOpPatcher from './CompoundAssignOpPatcher.js';
 
 export default class LogicalOpCompoundAssignOpPatcher extends CompoundAssignOpPatcher {
-  patch() {
+  patchAsExpression() {
     let operator = this.getOperatorToken();
     // `a &&= b` → `a && b`
     //      ^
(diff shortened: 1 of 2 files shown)


src/patchers/LogicalOpCompoundAssignOpPatcher.js
```diff
@@ -15,4 +15,28 @@ export default class LogicalOpCompoundAssignOpPatcher extends CompoundAssignOpPa
...
+
+    // `a &&= b` → `if (a &&= b`
+    //              ^^^^
+    this.insertBefore('if (');
+
+    if (op === '||=') {
+      this.assignee.negate();
+    }
+
+    let assigneeAgain = this.assignee.makeRepeatable(false);
```
jeffchan: Question: What does `makeRepeatable` do?

jeffchan:
:+1:

src/patchers/LogicalOpCompoundAssignOpPatcher.js
eventualbuddha: This is definitely a topic for CONTRIBUTING.md, but here's the basic summary. Most code is not safely repeatable because it may have side-effects. That is, in this case we have to expand `a &&= b` because the `&&=` operator doesn't exist in JavaScript. That means we build `if (a) { a = b; }`. This is fine since typically referencing an identifier doesn't usually have side-effects (but it could! *sigh* CoffeeScript decided that it's "safe", so we follow suit). However, in `a().b &&= c` it is clearly not safe to expand it to `if (a().b) { a().b = c; }` since `a()` would be called twice. So `makeRepeatable` transforms the node `a().b` to store the side-effecty parts and return some code that will access the same value, but won't trigger the side effects. In this case it'd become e.g. `if ((base = a()).b) { base.b = c; }`. So `makeRepeatable` edited the original node to be `(base = a()).b` and returned `base.b` (as `assigneeAgain`, in this case). Does that make sense?

jeffchan: Ah yes! That's clever. Hadn't thought of this edge case.

PR #134, opened 2016-02-28
WIP Add SwitchPatcher and SwitchCasePatcher

jeffchan:
WIP. Posting here to get feedback / letting others know I'm working on this.

### DIFF
--- a/src/patchers/SwitchCasePatcher.js
+++ b/src/patchers/SwitchCasePatcher.js
@@ -1,0 +1,95 @@
+import NodePatcher from './NodePatcher.js';
+import type { Editor, Node, ParseContext, Token } from './types.js';
+
+export default class SwitchCasePatcher extends NodePatcher {
+  constructor(node: Node, context: ParseContext, editor: Editor, conditions: NodePatcher, consequent: NodePatcher) {
+    super(node, context, editor);
+    this.conditions = conditions;
+    this.consequent = consequent;
+  }
+
+  patchAsStatement() {
+    // `when a, b, c then d` → `a, b, c then d`
+    //  ^^^^^
+    let whenToken = this.getWhenToken();
+    this.remove(whenToken.range[0], this.conditions[0].start);
+
+    // `a, b, c then d` → `a b c then d`
+    //   ^  ^
+    this.getCommaTokens().forEach((comma) => {
+      this.remove(...comma.range);
+    });
+
+    this.conditions.forEach((condition) => {
+      // `a b c then d` → `case a: case b: case c: then d`
+      //                   ^^^^^ ^^^^^^^ ^^^^^^^ ^
+      condition.insertBefore('case ');
+      condition.patch({ leftBrace: false, rightBrace: false });
+      condition.insertAfter(':');
+    });
+
+
+    // `case a: case b: case c: then d → `case a: case b: case c: d`
+    //                          ^^^^^
+    let thenToken = this.getThenToken();
+    if (thenToken) {
+      this.remove(thenToken.range[0], this.consequent.start);
+    }
+
+    this.consequent.patch({ leftBrace: false, rightBrace: false });
+
+    if (!this.getBreakToken() && !this.implicitlyReturns()) {
+      if (thenToken) {
+        // `case a: case b: case c: then d → `case a: case b: case c: d break`
+        //                                                             ^^^^^^
+        this.consequent.insertAfter(' break');
+      } else {
+        this.appendLineAfter('break', 1);
+      }
+    }
+  }
+
+  setImplicitlyReturns() {
+    super.setImplicitlyReturns();
+    this.consequent.setImplicitlyReturns();
+  }
+
+  patchAsExpression() {
+    this.patchAsStatement();
+  }
+
+  negate() {
+    this.conditions.forEach(condition => condition.negate());
+  }
+
+  /**
+   * @private
+   */
+  getWhenToken(): ?Token {
+    return this.tokens.find(token => token.type === 'LEADING_WHEN');
+  }
+
+  /**
+   * @private
+   */
+  getCommaTokens(): Array<Token> {
+    return this.tokens.filter(token => token.type === ',');
+  }
+
+  /**
+   * @private
+   */
+  getBreakToken(): ?Token {
+    return this.tokens.find(token => token.type === 'STATEMENT' && token.data === 'break');
+  }
+
+  /**
+   * Gets the token representing the `then` between condition and consequent.
+   *
+   * @private
+   */
+  getThenToken(): ?Token {
+    return this.tokenBetweenPatchersMatching(this.conditions[0], this.consequent, 'THEN');
+  }
+
+}
--- a/src/patchers/SwitchPatcher.js
+++ b/src/patchers/SwitchPatcher.js
@@ -1,0 +1,101 @@
+import NodePatcher from './NodePatcher.js';
+import type { Editor, Node, ParseContext, Token } from './types.js';
+
+export default class SwitchPatcher extends NodePatcher {
+  constructor(node: Node, context: ParseContext, editor: Editor, expression: NodePatcher, cases: Array<NodePatcher>, alternate: ?NodePatcher) {
+    super(node, context, editor);
+    this.expression = expression;
+    this.cases = cases;
+    this.alternate = alternate;
+  }
+
+  patchAsStatement() {
+    if (this.expression) {
+      // `switch a` → `switch (a`
+      //                      ^
+      if (!this.expression.isSurroundedByParentheses()) {
+        this.expression.insertBefore('(');
+      }
+
+      this.expression.patch();
+
+      // `switch (a` → `switch (a)`
+      //                         ^
+      if (!this.expression.isSurroundedByParentheses()) {
+        this.expression.insertAfter(')');
+      }
+
+      // `switch (a)` → `switch (a) {`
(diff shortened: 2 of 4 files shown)


src/patchers/SwitchPatcher.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  patchAsExpression() {
+    this.patchAsStatement();
+  }
+
+  /**
+   * @private
+   */
+  overwiteElse() {
```
eventualbuddha: Typo

src/patchers/SwitchCasePatcher.js
```diff
@@ -0,0 +1,86 @@
...
+    let { consequent } = this;
+    consequent.setImplicitlyReturns();
+  }
+
+  patchAsExpression() {
+    this.patchAsStatement();
+  }
+
+  negate() {
+    this.negated = !this.negated;
```
eventualbuddha: This should negate the conditions.

src/patchers/SwitchCasePatcher.js
```diff
@@ -0,0 +1,86 @@
...
+    let whenToken = this.getWhenToken();
+    this.remove(whenToken.range[0], conditions[0].start);
+
+    this.getCommaTokens().forEach((comma) => {
+      this.remove(...comma.range);
+    });
+
+    conditions.forEach((condition) => {
+      condition.insertBefore('case ');
+      if (this.negated) {
```
eventualbuddha: This is not needed if `negate` is fixed.

src/patchers/SwitchCasePatcher.js
```diff
@@ -0,0 +1,86 @@
...
+    });
+
+    let thenToken = this.getThenToken();
+    if (thenToken) {
+      this.remove(thenToken.range[0], consequent.start);
+    }
+
+    consequent.patch({ leftBrace: false, rightBrace: false });
+
+    // TODO: Only insert break when necessary
```
eventualbuddha: This could be don't as an option passed to `patch` in `SwitchPatcher`.

src/patchers/SwitchPatcher.js
```diff
@@ -0,0 +1,83 @@
...
+
+    if (expression) {
+      if (!expression.isSurroundedByParentheses()) {
+        expression.insertBefore('(');
+        expression.patch();
+        expression.insertAfter(')');
+      }
+    } else {
+      this.cases.forEach((casePatcher) => casePatcher.negate());
+      this.insert(endOfLine, ' (false)');
```
eventualbuddha: maybe just get the `switch` token instead of finding the end of the line? What if it looks like this?

```coffee
switch # test
  when a
```

src/patchers/SwitchPatcher.js
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    this.cases.forEach((casePatcher) => casePatcher.patch());
```
eventualbuddha: I should add an eslint rule for it, but I typically omit parentheses when there's a single argument.

src/patchers/SwitchCasePatcher.js
```diff
@@ -0,0 +1,86 @@
...
+export default class SwitchCasePatcher extends NodePatcher {
+  constructor(node: Node, context: ParseContext, editor: Editor, conditions: NodePatcher, consequent: NodePatcher) {
+    super(node, context, editor);
+    this.conditions = conditions;
+    this.consequent = consequent;
+    this.negated = false;
+  }
+
+  patchAsStatement() {
+    let { conditions, consequent } = this;
```
eventualbuddha: I've been moving away from this pattern and just accessing properties on `this` as needed.

src/patchers/SwitchPatcher.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  patchAsStatement() {
+    let { expression, alternate } = this;
+    let endOfLine = this.getFirstLineBreak();
+
+    if (expression) {
+      if (!expression.isSurroundedByParentheses()) {
+        expression.insertBefore('(');
+        expression.patch();
```
eventualbuddha: The expression won't be patched if it is surrounded by parentheses.

src/patchers/SwitchCasePatcher.js
jeffchan: I can do that. I did it this way to match the expected test output https://github.com/decaffeinate/decaffeinate/blob/2a185dcc8cec0a7a665b23d33212861ba7b18fc1/test/switch_test.js#L170-L186 (which is also the same as the CoffeeScript output). If we negate the condition, then our EqualityPatcher will flip signs. That's okay, right?

eventualbuddha: Yep, I think that's fine. I'm not aware of a situation where that could cause issues.

test/switch_test.js
```diff
@@ -161,9 +210,8 @@ describe.skip('switch', () => {
           case c:
             return d;
         }
-      };
-
-      e;
+      }
+      ;e;
```
jeffchan: Need some help figuring out this one.

These are the patcher edits:
```
FunctionPatcher INSERT 4 "function" BEFORE "->"
FunctionPatcher INSERT 4 "() " BEFORE "->"
FunctionPatcher OVERWRITE [4, 6) "->" → "{"
IdentifierPatcher INSERT 16 "(" BEFORE "b\n"
IdentifierPatcher INSERT 17 ")" BEFORE "\n "
IdentifierPatcher INSERT 17 " {" BEFORE "\n "
SwitchCasePatcher REMOVE [22, 27) "when "
IdentifierPatcher INSERT 27 "case " BEFORE "c\n"
IdentifierPatcher INSERT 28 ":" BEFORE "\n "
BlockPatcher INSERT 35 "return " BEFORE "d\n"
BlockPatcher INSERT 36 ";" BEFORE "\ne"
SwitchPatcher INSERT 36 "\n  }" BEFORE "\ne"
BlockPatcher INSERT 37 ";" BEFORE "e"  <---
BlockPatcher INSERT 36 "\n}" BEFORE "\ne" <---
BlockPatcher INSERT 37 ";" BEFORE "e"  <---
BlockPatcher INSERT 38 ";" BEFORE ""
```

It looks like the BlockPatcher semicolon insertion is conflicting with the right brace insertion.

test/switch_test.js
```diff
@@ -177,16 +225,16 @@ describe.skip('switch', () => {
...
+        case score >= 70: 'D'; break;
+        case score >= 80: 'C'; break;
+        case score >= 90: 'B'; break;
         default: 'A';
       }
     `);
   });
 
-  it('works with `switch` used as an expression', () => {
+  it.skip('works with `switch` used as an expression', () => {
```
jeffchan: This one is close, but I can't figure out how to indent :confused: 

```
  1) switch works with `switch` used as an expression:

      AssertionError: 'var a = (function() {\nswitch (b) {\n  case c: return d;\n  default: return e;\n}\n})();' === 'var a = (function() {\n  switch (b) {\n    case c: return d;\n    default: return e;\n  }\n})();'
      + expected - actual

       var a = (function() {
      -switch (b) {
      -  case c: return d;
      -  default: return e;
      -}
      +  switch (b) {
      +    case c: return d;
      +    default: return e;
      +  }
       })();
```

test/switch_test.js
```diff
@@ -201,7 +249,7 @@ describe.skip('switch', () => {
     `);
   });
 
-  it('works with the switch from the CoffeeScript demo page', () => {
+  it.skip('works with the switch from the CoffeeScript demo page', () => {
```
jeffchan: Trying to debug this one. Looks like my `break` insertion is off by one line

```
  2) switch works with the switch from the CoffeeScript demo page:

      AssertionError: 'switch (day) {\n  case "Mon": go(work); break;\n  case "Tue": go(relax); break;\n  case "Thu": go(iceFishing); break;\n  case " === 'switch (day) {\n  case "Mon": go(work); break;\n  case "Tue": go(relax); break;\n  case "Thu": go(iceFishing); break;\n  case "
      + expected - actual

           if (day === bingoDay) {
             go(bingo);
             go(dancing);
           }
      -  case "Sun": go(church);
      -    break; break;
      +    break;
      +  case "Sun": go(church); break;
         default: go(work);
       }
```

test/switch_test.js
eventualbuddha: I'll take a look.

test/switch_test.js
eventualbuddha: @jeffchan You can now use `NodePatcher#indent()` (added in https://github.com/decaffeinate/decaffeinate/commit/8e2b67698f3f78ca77da9c9f1123ffeeb99879cc).

test/switch_test.js
eventualbuddha: Usually this has to do with the editing not happening in the order you expect. I'll take a look today if I get a chance.

test/switch_test.js
eventualbuddha: This is due to the way CoffeeScript encodes location information about functions. Oddly, it does the same thing CoffeeScriptRedux does where they consume as much non-semantic source as possible after the end of the function and include it in the function's range, so the placement of the semicolons afterward is wrong. Don't worry about this, it needs to be fixed in decaffeinate-parser.

eventualbuddha:
Superceded by https://github.com/decaffeinate/decaffeinate/pull/137.

PR #137, opened 2016-03-04
Add support for `switch`

eventualbuddha:
This is a work in progress. There are still a few pending tests that need fixing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,8 +31,8 @@
   "license": "MIT",
   "dependencies": {
     "add-variable-declarations": "^1.2.0",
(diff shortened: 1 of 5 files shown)


eventualbuddha:
@jeffchan you can continue working on this branch once you accept the organization invite.

eventualbuddha:
The problem with the `break` being inserted in the wrong place is due to the same location info bug I mentioned before where it eats all the non-semantic source after the closing of the block and considers it part of the node.

eventualbuddha:
@jeffchan I believe this all works now. Let me know what you think.

PR #140, opened 2016-03-14
Add esnext as the final transformation stage.

eventualbuddha:
This upgrades the resulting code to use even more modern features, such as `let`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,10 +31,12 @@
   "license": "MIT",
   "dependencies": {
     "add-variable-declarations": "^1.2.0",
+    "babel-eslint": "^5.0.0",
     "coffee-lex": "^1.3.2",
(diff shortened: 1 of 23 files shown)


jeffchan:
Sweet. I've been doing this via a script. One issue is that if the coffee looks something like:
```coffee
render: ->
  `<MyJsxComponent />`
```

Then the transformation will fail at the esnext stage since esnext doesn't support JSX out of the box (see https://github.com/esnext/esnext/pull/88)

eventualbuddha:
@jeffchan I just [added a test](https://github.com/decaffeinate/decaffeinate/commit/8243c538723fb2a20f96cd76c030be432a18ad5b) for JSX on this branch.

jeffchan:
Thanks!

PR #145, opened 2016-03-17
Ensure conditional-as-ternary has parens if needed.

### LINKED ISSUE #144: Ternary Operator Implies Parenthesis
In the [contribution guide](https://github.com/decaffeinate/decaffeinate/blob/master/CONTRIBUTING.md#patchers-with-children), you wrote:
>Now `a + if b then c else d` will become `a + b ? c : d;` as intended, since the right patcher knows to patch using patchAsExpression instead of patchAsStatement. In reality, PlusOp and several other binary operators are all handled by BinaryOpPatcher.`

Shouldn't this more correctly be patched as `a + (b ? c : d)`? 

Because of the condition `b` is enclosed within the ternary, parenthesis are implied here (and generated by the coffeescript compiler in this case). We probably shouldn't always add parenthesis, but I believe they should become explicit/required as a general case whenever a left-associative operator meets a right-associative operator.

eventualbuddha:
Fixes #144.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -326,8 +326,8 @@ export default class PlusOpPatcher extends NodePatcher {
 }
 ```
 
-Now `a + if b then c else d` will become `a + b ? c : d;` as intended, since the
-`right` patcher knows to patch using `patchAsExpression` instead of
+Now `a + if b then c else d` will become `a + (b ? c : d);` as intended, since
+the `right` patcher knows to patch using `patchAsExpression` instead of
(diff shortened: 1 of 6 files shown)


eventualbuddha:
@thebanjomatic any other cases you can think of besides binary and unary operators?

eventualbuddha:
Okay, I changed how we decide to add parens to make your new example work and added a test for it. Anything else?

thebanjomatic:
That looks great, thanks for the fast turnaround!

PR #151, opened 2016-03-18
Add support for unit testing and test `BlockPatcher`.

eventualbuddha:


eventualbuddha:
Something about the importing of the source files directly instead of the built `decaffeinate.cjs.js` is causing babel to behave strangely. I just committed the changes to `BlockPatcher` directly so this does not apply anymore. Maybe I'll revisit later.

PR #183, opened 2016-04-30
Converted ESLint rules into a plugin

flying-sheep:
as rulesdir is deprecated: eslint/eslint#2715

this has the sideeffect that you can run eslint without commandline options


eventualbuddha:
Thanks!

flying-sheep:
np!

PR #184, opened 2016-04-30
Added mocha.opts

### LINKED ISSUE #182: consider using mocha.opts
you can put a file containing them into the `tests` directory, then mocha runners like atom’s mocha-ui will do the right thing.

```bash
--compilers js:./test/support/babel-register.js
--recursive
```

it’s weird that the path is relative from the project dir instead of `tests/`, though.

flying-sheep:
fixes #182


eventualbuddha:
Good to know. Thanks!

PR #187, opened 2016-05-01
Assign parameters to properties

marcmenn:
support thinks like `constructor: (@a) ->`

### DIFF
--- a/src/stages/normalize/patchers/DefaultParamPatcher.js
+++ b/src/stages/normalize/patchers/DefaultParamPatcher.js
@@ -1,0 +1,3 @@
+import PassthroughPatcher from '../../../patchers/PassthroughPatcher.js'
+
+export default class DefaultParamPatcher extends PassthroughPatcher {}
--- a/src/stages/normalize/patchers/DefaultParamPatcher.js
+++ b/src/stages/normalize/patchers/DefaultParamPatcher.js
@@ -1,3 +1,3 @@
-import PassthroughPatcher from '../../../patchers/PassthroughPatcher.js'
+import PassthroughPatcher from '../../../patchers/PassthroughPatcher.js';
 
 export default class DefaultParamPatcher extends PassthroughPatcher {}
--- a/src/stages/normalize/patchers/FunctionPatcher.js
+++ b/src/stages/normalize/patchers/FunctionPatcher.js
@@ -1,9 +1,9 @@
 import NodePatcher from './../../../patchers/NodePatcher.js';
-import type { Node, ParseContext, Editor, SourceToken } from './../../../patchers/types.js';
+import type { Node, ParseContext, Editor } from './../../../patchers/types.js';
 
 export default class FunctionPatcher extends NodePatcher {
   parameters: Array<NodePatcher>;
-  body: ?BlockPatcher;
+  body: ?NodePatcher;
   
   constructor(node: Node, context: ParseContext, editor: Editor, parameters: Array<NodePatcher>, body: ?NodePatcher) {
     super(node, context, editor);
@@ -17,7 +17,7 @@ export default class FunctionPatcher extends NodePatcher {
     // This is detected and used by the MemberAccessOpPatcher to claim a free binding for this parameter
     // (from the functions scope, not the body's scope)
 
-    const assignments = [];
+    let assignments = [];
     this.node._assignMember = function(memberName: string){
       let varName = this.claimFreeBinding(memberName);
       assignments.push(`@${memberName} = ${varName}`);
--- a/src/stages/normalize/patchers/FunctionPatcher.js
+++ b/src/stages/normalize/patchers/FunctionPatcher.js
@@ -18,16 +18,16 @@ export default class FunctionPatcher extends NodePatcher {
     // (from the functions scope, not the body's scope)
 
     let assignments = [];
-    this.node._assignMember = function(memberName: string){
+    this.addStatementAtScopeHeader = (memberName: string) => {
       let varName = this.claimFreeBinding(memberName);
       assignments.push(`@${memberName} = ${varName}`);
       this.log(`Replacing parameter @${memberName} with ${varName}`);
       return varName;
-    }.bind(this);
+    };
 
     this.parameters.forEach(parameter => parameter.patch());
 
-    delete this.node._assignMember;
+    delete this.addStatementAtScopeHeader;
 
     // If there were assignments from parameters insert them
     if (this.body) {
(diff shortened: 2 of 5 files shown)


src/stages/normalize/patchers/FunctionPatcher.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  patchAsExpression() {
+    // To avoid knowledge of all the details how assignments can be nested in nodes,
+    // we add a callback to the function node before patching the parameters and remove it afterwards.
+    // This is detected and used by the MemberAccessOpPatcher to claim a free binding for this parameter
+    // (from the functions scope, not the body's scope)
+
+    let assignments = [];
+    this.node._assignMember = function(memberName: string){
```
marcmenn: this *is* ugly, might work nicer, if patch is a generator function, so there could be easier communication between caller and nested callee…

eventualbuddha: Yeah. I'd prefer to keep the nodes as they are and communicate between patchers via the `parent` property. Perhaps something like a `addStatementAtScopeHeader`?

src/stages/normalize/patchers/FunctionPatcher.js
```diff
@@ -0,0 +1,52 @@
...
+      let varName = this.claimFreeBinding(memberName);
+      assignments.push(`@${memberName} = ${varName}`);
+      this.log(`Replacing parameter @${memberName} with ${varName}`);
+      return varName;
+    }.bind(this);
```
eventualbuddha: We're not keeping this method, but if we were this could be an arrow function instead of a `.bind(this)` function.

src/stages/normalize/patchers/MemberAccessOpPatcher.js
```diff
@@ -0,0 +1,25 @@
+import PassthroughPatcher from '../../../patchers/PassthroughPatcher.js';
+
+export default class MemberAccessOpPatcher extends PassthroughPatcher {
+  shouldTrimContentRange() {
+    return true;
+  }
+
+  patch() {
+    super.patch();
+    let cb = this.findAssignMemberCallback();
```
eventualbuddha: Let's use the longer `callback` instead of `cb`.

test/class_test.js
```diff
@@ -93,6 +93,67 @@ describe('classes', () => {
...
+          method: (@a) =>
+      `, `
+        class A {
+          constructor() {
+            this.method = this.method.bind(this);
+          }
+        
+          method(a) {
+            this.a = a;
+            return;
```
eventualbuddha: This is unfortunate since it's not the way a person would have written it. Is doing this transform during the main stage too complicated?

test/class_test.js
```diff
@@ -93,6 +93,67 @@ describe('classes', () => {
...
+      `, `
+        class A {
+          method(a) {
+            this.a = a;
+            return;
+          }
+        }
+      `);
+    });
+    it('bound method', () => {
```
eventualbuddha: nit: empty line between `it` tests.

test/class_test.js
```diff
@@ -93,6 +93,67 @@ describe('classes', () => {
     `);
   });
 
+  describe('assign properties from method parameters', () => {
```
eventualbuddha: These tests look good! A few that we should add:

```js
it('chooses variables that do not conflict', () => {
  check(`
    (@a) ->
      a = 1
  `, `
    (function(a$1) {
      this.a = a$1;
      let a;
      return a = 1;
    });
  `);
});

it('prepends assignments in single-line methods', () => {
  check(`
    class A
      method: (@a) -> console.log(@a)
  `, `
    class A {
      method(a) { this.a = a; console.log(this.a); }
    }
  `);
});
```

test/class_test.js
eventualbuddha: Another possibility is to use an identifier that we're pretty sure won't be actually returned. So use `return DECAFFEINATE_24820832` where the number is calculated once using `Date.now()` or something, then the main stage could look for such return statements and remove them.

marcmenn: We often use literal `return` statements in coffee to avoid for example getting `for` loops as expressions. So I feel this a natural thing to do. On the other hand, we could generally drop `return` statements (without values) when they are the last statement in a function? Maybe it's not necessary to introduce such a magic identifier…

eventualbuddha: That's a good point. Tracked as #193.

src/stages/normalize/patchers/FunctionPatcher.js
```diff
@@ -0,0 +1,57 @@
...
+    this.parameters.forEach(parameter => parameter.patch());
+
+    delete this.addStatementAtScopeHeader;
+
+    // If there were assignments from parameters insert them
+    if (this.body) {
+      // before the actual body
+      if (assignments.length) {
+        let text;
+        if (this.body.node.inline) {
```
eventualbuddha: I think all this can be replaced with calls to `this.body.insertStatementsAtIndex`.

marcmenn: Which lives in the main stage… I ran into trouble trying all this in the main stage with the editor complaining about editing positions twice, if I remember right. Maybe this was, because the actual parameter (e.g. `@a`) could be processed in different ways from other patchers…

eventualbuddha: Ah, right. Okay, this is fine for now but I'll make a note to extract that functionality into a shared `BlockPatcher` that both stages inherit from.

src/stages/normalize/patchers/DefaultParamPatcher.js
```diff
@@ -0,0 +1,3 @@
+import PassthroughPatcher from '../../../patchers/PassthroughPatcher.js';
+
+export default class DefaultParamPatcher extends PassthroughPatcher {}
```
eventualbuddha: This isn't needed, right?

eventualbuddha: Ah, I see, you're checking for it below.

eventualbuddha:
So this is ready to go, right? No other cleanup you want to do ?

marcmenn:
Yes, think it's ready. Thanks for your comments!

marcmenn:
Found another lovely [edgecase](http://coffeescript.org/#try:(%40a%2C%20%40b%20%3D%20%40a)%20-%3E%0A%0A(%40a%2C%20b%20%3D%20%40a)%20-%3E%0A%20%20%40a%20%3D%20a%0A%20%20return) which is not handled correctly yet. Added a skipped test for that in 08fccde0a90aea73babf28144b617f38e9b4d16c

eventualbuddha:
Thanks @marcmenn! Released in v2.8.0.

Issue #125, opened 2016-02-16
Paramter defaults for passed-in null values are different in coffeescript and es6 


jcalfee:
Null passed in coffee-script resolves as to the parameter's default value but resolves as `null` in es6.  Only `undefined` in es6 will capture a default value.

``` coffeescript
    Aes.decrypt_with_checksum = (private_key, public_key, nonce = "", message) ->
```

should convert to

``` js
Aes.decrypt_with_checksum = function(private_key, public_key, nonce, message) {
  if (nonce == null) {
    nonce = "";
  }
}
```

This will not work:

``` js
    Aes.decrypt_with_checksum = function(private_key, public_key, nonce = "", message) {
    };

```

eventualbuddha:
Ugh, that's sad. Probably best to make this an option.

an-sh:
Sorry for the duplicate. The best option for me: for any code by
default it should produce semantically 100% equivalent or fail with an
error otherwise. So the first is rewriting the coffee (for missing in
es6 itself or this convertor features), then compilation will give
100% identical functionally. Better missing some constructs support,
that an incorrect output. In this work-flow model no debugging of es5
output is required.

For this case, probably it will be good to have some flag (default
disabled) that will enable es6 default params usage.

alangpierce:
For a first pass, I'll just implement the conditional assignment code without leaving in the default or adding the comment. Those can be added later if we think they're useful.

PR #188, opened 2016-05-01
Update README.md

coderofsalvation:
*'time to move on'* might look a bit opiniated / demotivating to use coffeescript as a syntax.
Imho coffeescript users can just keep using coffeescript syntax thanks to decaffeinate :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,7 @@
 # decaffeinate [![Build Status](https://travis-ci.org/decaffeinate/decaffeinate.svg?branch=master)](https://travis-ci.org/decaffeinate/decaffeinate)
 
 CoffeeScript in, modern JavaScript out.
-
-JavaScript is the future, in part thanks to CoffeeScript. Now that it has served
-its purpose, it's time to move on. Convert your CoffeeScript source to modern
-JavaScript with decaffeinate.
+Convert your CoffeeScript source to modern JavaScript with decaffeinate.
 
 ## Install
 


hpurmann:
Admitted, it's pretty opinionated. But by now it's almost clear that CoffeeScript will die sooner or later. It's a pretty bad idea to start a new project/module using it. This tool really helps to accelerate the conversion process.

So in conclusion, I'm against merging this.

coderofsalvation:
Thanks for your reply.
You are the maintainer, so ofcourse you are entitled to promote any opinion you like.
However, why do you think its bad to write a module with coffeescript (in `src` and the transpiled code in `lib`) besides less beautiful commit-logs?

hpurmann:
No, I'm not the maintainer of this project. I just gave my opinion.

Why I think that ES6 is the future? First, it's the new javascript standard. You can expect browsers and nodejs to work towards supporting/implementing this standard. 
Second, the tooling. For example take JSX, Facebook Flow or Code Coverage tools. The community decided to use ES6 and that's why these and upcoming tools are supporting it.
And lastly, where is the need for Coffeescript nowadays? Most of the syntactic sugar Coffeescript has are now in ES6. Coffee was good for its time but now it's time to move on.

coderofsalvation:
Hm..I don't really get this ES6 vs coffeescript apple-vs-pears thing.
Why would one of them become 'the future', and why should a community bother being against something?
However, I **can** see the benefits of purely es6 modules though, backwards transpilable with older es-versions, removes the compilation to js-step, and promotes single-dialect modules.
Anyways, thanks for the feedback I definately respect and understand your point of view.

eventualbuddha:
Thanks for your contribution, @coderofsalvation! I agree the language of the README is strong, and moderating it may be a good idea. That said, the reasons this project exists are mostly laid out by @hpurmann:

* ES6 (and ES7 etc) are the future of JavaScript, and won't need a translation layer, so users of ES6 do not need to learn another dialect to write JavaScript
* Tooling is poor in CoffeeScript. Yes, there exists syntax highlighters and basic versions of linters and code coverage tools etc, but those tools are always poor compared to the equivalents in JavaScript simply because there are not enough users who spend the time required to bring the quality up to par. Just look at [this 4 year-old ticket to see what I mean](https://github.com/clutchski/coffeelint/issues/20).
* Perhaps most important, CoffeeScript represented a JavaScript experiment which turned out to be a great success. ES6 is heavily influenced by the feature set of CoffeeScript. However, CoffeeScript is no longer where the bulk of JavaScript experimentation is happening. Babel is.

Without the experimentation that made CoffeeScript such a success, and without quality tools, I'd argue that no one today should be starting a new project in CoffeeScript. It is an evolutionary dead-end and this project was primarily developed so that Square (my employer) could transition off of it with relative ease.

Finally, the goals of this project are different than that of the official CoffeeScript project. I make a strong effort to produce code that will behave the same as the original CoffeeScript code *most of the time*, but favor ES6 equivalents even if there are slight differences in semantics (see #125). That, and it's probably slow compared to the official compiler. That means this project should not be relied upon as a drop-in replacement for the official CoffeeScript compiler.

If updating the README with any of this information would be useful to you, I'd happily review a PR 😉

coderofsalvation:
"That means this project should not be relied upon as a drop-in replacement for the official CoffeeScript compiler"

PR #190, opened 2016-05-04
yield/generator support

marcmenn:
Based on the recent support of yield in coffee-lex and decaffeinate-parser. Two new Patchers, one for `yield` and one for `yield from`.


src/stages/main/patchers/GeneratorFunctionPatcher.js
```diff
@@ -0,0 +1,24 @@
+import FunctionPatcher from './FunctionPatcher.js';
+
+/**
+ * Handles bound functions, i.e. "fat arrows".
```
eventualbuddha: Copy-paste error.

src/stages/main/patchers/GeneratorFunctionPatcher.js
```diff
@@ -0,0 +1,24 @@
...
+ */
+export default class GeneratorFunctionPatcher extends FunctionPatcher {
+
+  patchFunctionStart({ method=false }) {
+    this.log('Running with method=' + method);
```
eventualbuddha: Probably don't need to leave this in.

src/stages/main/patchers/GeneratorFunctionPatcher.js
```diff
@@ -0,0 +1,24 @@
...
+    }
+    
+    if (!this.hasParamStart()) {
+      this.insert(this.contentStart, '() ');
+    }
+
+    this.overwrite(arrow.start, arrow.end, '{');
+  }
+
+
```
eventualbuddha: Trim extra lines please.

src/stages/main/patchers/YieldFromPatcher.js
```diff
@@ -0,0 +1,22 @@
...
+export default class YieldFromPatcher extends NodePatcher {
+  expression: NodePatcher;
+  
+  constructor(node: Node, context: ParseContext, editor: Editor, expression: NodePatcher) {
+    super(node, context, editor);
+    this.expression = expression;
+  }
+
+  /**
+   * YIELD FROM EXPRESSION
```
eventualbuddha: Since the `yield from` part is literal text, use `'yield' 'from' EXPRESSION`.

src/utils/traverse.js
```diff
@@ -103,6 +103,9 @@ const ORDER = {
   RemOp: ['left', 'right'],
   Rest: ['expression'],
   Return: ['expression'],
+  GeneratorFunction: ['parameters', 'body'],
+  Yield: ['expression'],
+  YieldFrom: ['expression'],
```
eventualbuddha: Please sort alphabetically.

src/stages/main/patchers/YieldPatcher.js
```diff
@@ -0,0 +1,19 @@
...
+export default class YieldPatcher extends NodePatcher {
+  expression: NodePatcher;
+  
+  constructor(node: Node, context: ParseContext, editor: Editor, expression: NodePatcher) {
+    super(node, context, editor);
+    this.expression = expression;
+  }
+
+  /**
+   * YIELD EXPRESSION
```
eventualbuddha: Same as the other, `'yield' EXPRESSION`.

src/stages/main/patchers/WhilePatcher.js
```diff
@@ -125,9 +125,21 @@ export default class WhilePatcher extends NodePatcher {
...
+    let prefix = isYielding ? 'yield* (function*()' : '(() =>';
+    this.insert(this.contentStart, `${prefix} { ${resultBinding} = []; `);
+    let postfix = isYielding ? '.call(this)' : '()';
+    this.insert(this.contentEnd, ` return ${resultBinding}; })${postfix}`);
+    if (isYielding) {
+      this.yields();
+    }
+  }
+
+  yieldController(node) {
```
eventualbuddha: Missing type annotation (i.e. `node: Node`). But why do we even need the node?

src/stages/main/patchers/YieldPatcher.js
```diff
@@ -0,0 +1,19 @@
...
+   */
+  patchAsExpression({ needsParens=true }={}) {
+    this.expression.patch({ needsParens });
+    this.yields();
```
eventualbuddha: Can this not happen during `initialize()`, similar to `return` handling?

src/stages/main/patchers/YieldFromPatcher.js
```diff
@@ -0,0 +1,22 @@
...
+   */
+  patchAsExpression({ needsParens=true }={}) {
+    let src = this.sourceTokenAtIndex(this.contentStartTokenIndex);
+    this.overwrite(src.start, src.end, 'yield*');
+    
+    this.expression.patch({ needsParens });
+    this.yields();
```
eventualbuddha: Ditto on `initialize`?

src/stages/main/patchers/YieldFromPatcher.js
```diff
@@ -0,0 +1,22 @@
...
+  constructor(node: Node, context: ParseContext, editor: Editor, expression: NodePatcher) {
+    super(node, context, editor);
+    this.expression = expression;
+  }
+
+  /**
+   * YIELD FROM EXPRESSION
+   */
+  patchAsExpression({ needsParens=true }={}) {
+    let src = this.sourceTokenAtIndex(this.contentStartTokenIndex);
```
eventualbuddha: We can just overwrite from `this.contentStart` to `this.contentEnd`, there's no need to get the underlying token.

src/utils/parse.js
```diff
@@ -20,6 +20,7 @@ function attachScope(node: Node) {
 
     case 'Function':
     case 'BoundFunction':
+    case 'GeneratorFunction':
```
eventualbuddha: This makes me realize I have some cleanup to do. You had to add this in too many places 😞

src/stages/main/patchers/GeneratorFunctionPatcher.js
```diff
@@ -0,0 +1,24 @@
+import FunctionPatcher from './FunctionPatcher.js';
+
+/**
+ * Handles bound functions, i.e. "fat arrows".
+ */
+export default class GeneratorFunctionPatcher extends FunctionPatcher {
+
```
eventualbuddha: Trim extra line please.

src/stages/main/patchers/YieldFromPatcher.js
marcmenn: Well, actually not, I don't understand to much about the code, but in the test contentStart/contentEnd also overwrites the (elsewhere) added `return` statement, so I had to go through tokens

src/stages/main/patchers/WhilePatcher.js
marcmenn: Well, because of… no… not really. don't need this.

src/patchers/NodePatcher.js
```diff
@@ -869,4 +869,14 @@ export default class NodePatcher {
   allCodePathsPresent(): boolean {
     return true;
   }
+
+  yields() {
+    let receiver = this.parent;
+    while (receiver) {
+      if (receiver.yieldController) {
+        return receiver.yieldController(this);
```
eventualbuddha: This passes an argument but the method doesn't expect one.

src/stages/main/patchers/WhilePatcher.js
```diff
@@ -125,9 +126,16 @@ export default class WhilePatcher extends NodePatcher {
   patchAsExpression() {
     this.body.setImplicitlyReturns();
     let resultBinding = this.getResultArrayBinding();
-    this.insert(this.contentStart, `(() => { ${resultBinding} = []; `);
     this.patchAsStatement();
-    this.insert(this.contentEnd, ` return ${resultBinding}; })()`);
+    let prefix = this.yielding ? 'yield* (function*()' : '(() =>';
+    this.insert(this.contentStart, `${prefix} { ${resultBinding} = []; `);
+    let postfix = this.yielding ? '.call(this)' : '()';
```
eventualbuddha: We should change `.call(this)` to `.apply(this, arguments)`, since the `while` body might reference `arguments`, which would refer to a different value inside a generator function.

marcmenn: We actually declare the function in 130 without arguments…

marcmenn: Coffee itself passes arguments _only_ when referenced inside body. So, if we support this, we might go for the same logic? Isn't it a rare case and produces unreadable code if arguments are not referenced? Also, it was not present in the logic before this PR and it disallows the nice and readable usage of the fat arrow. Would suggest to support this at a later time?

hpurmann:
I'd love to get this merged soon.

src/stages/main/patchers/WhilePatcher.js
eventualbuddha: I fixed this in a follow-up commit (https://github.com/decaffeinate/decaffeinate/commit/ac3b6434dc4500ef89c579aaccf1203105b73e26).

eventualbuddha:
Thanks @marcmenn! Released in v2.10.0 ([repl](http://decaffeinate.github.io/decaffeinate/repl/#?evaluate=true&code=-%3E%0A%20%20yield%201)).

hpurmann:
:+1:

PR #191, opened 2016-05-04
fix isSurroundedByParentheses

marcmenn:
current implementation assumes "surrounded", when it starts with a left and ends with a right parenthese, which is false positiv in case of `(a) and (b)`


eventualbuddha:
Thanks! Fixed in v2.7.10.

PR #218, opened 2016-05-14
Handle implicit objects wrapped in parens

iainb:
Some code that I've been attempting to convert to es6 does some fairly odd looking things with parens. However, it is valid CoffeeScript and when compiled to JavaScript "does the right thing".

For example:

```
someFunction = ->
    return (
        a: b
        b: c
    )
```
Compiles to:
```
var someFunction;

someFunction = function() {
  return {
    a: b,
    b: c
  };
};
```
In decaffeinate this currently results in an error raised in the ObjectInitialiserPatcher:

```
NormalizeStage someFunction = ->
    return (
        a: b
        b: c
    )

MainStage someFunction = ->
    return (
        a: b
        b: c
    )

FunctionPatcher INSERT 15 "function" BEFORE "->\n    r"
FunctionPatcher INSERT 15 "() " BEFORE "->\n    r"
FunctionPatcher OVERWRITE [15, 17) "->" → "{"
ObjectInitialiserPatcher INSERT 30 "{\n        " BEFORE "\n       "
ObjectInitialiserPatcher INSERT 29 "    " BEFORE "(\n      "
fail.coffee: cannot edit index 29 because it is not editable (i.e. outside [30, 61))
  1 | someFunction = ->
> 2 |     return (
> 3 |         a: b
> 4 |         b: c
> 5 |     )
  6 | 
```

This PR attempts to fix the above issue and 2 other closely related cases. I've broken each fix down into a single commit with fix + test to make reviewing a little easier. I'm happy to squash into a single commit if you want.

### DIFF
--- a/src/stages/main/patchers/ObjectInitialiserPatcher.js
+++ b/src/stages/main/patchers/ObjectInitialiserPatcher.js
@@ -29,7 +29,7 @@ export default class ObjectInitialiserPatcher extends NodePatcher {
       let textToInsert = '{';
       let shouldIndent = false;
       if (this.shouldExpandCurlyBraces()) {
-        if (this.implicitlyReturns()) {
+        if (this.implicitlyReturns() && !this.isSurroundedByParentheses()) {
           textToInsert = `{\n${this.getIndent()}`;
           shouldIndent = true;
         } else {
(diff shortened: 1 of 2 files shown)


iainb:
I don't think the travis-ci build failure is related to my changes, I successfully ran rm -rf node_modules; npm install; npm test without error on my machine (using node 6.1.0 / npm 3.8.6).

eventualbuddha:
Thanks!

eventualbuddha:
Released in v2.9.2.

PR #222, opened 2016-05-15
add content/location to error message from addVariableDeclarations

marcmenn:
addVariableDeclarations passes the generated JS through babel for the first time. if this throws an exception, you're left with pretty much no idea what happened.
now, if this exceptions contains `loc` information, the generated JS is appended to the err.message to give a hint, where decaffeinate failed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   "author": "Brian Donovan",
   "license": "MIT",
   "dependencies": {
-    "add-variable-declarations": "^1.5.0",
+    "add-variable-declarations": "1.5.0",
     "ast-processor-babylon-config": "^1.0.0",
     "automatic-semicolon-insertion": "^1.0.1",
     "babylon": "^6.7.0",
--- a/src/stages/add-variable-declarations/index.js
+++ b/src/stages/add-variable-declarations/index.js
@@ -9,7 +9,19 @@ export default class AddVariableDeclarationsStage {
     log(content);
 
     let editor = new MagicString(content);
-    addVariableDeclarations(content, editor);
+    try {
+      addVariableDeclarations(content, editor);
+    } catch (err) {
+      if (err.loc) {
+        let {line, column} = err.loc;
+        let lines = content.split('\n');
+        let l = '-'.repeat(column) + '^';
+        lines.splice(line, 0, l, '').join('\n');
+        content = lines.join('\n');
+      }
+      err.message += '\n' + content;
+      throw(err);
+    }
     return {
       code: editor.toString(),
       map: editor.generateMap({


package.json
```diff
@@ -30,7 +30,7 @@
   "author": "Brian Donovan",
   "license": "MIT",
   "dependencies": {
-    "add-variable-declarations": "^1.5.0",
+    "add-variable-declarations": "1.5.0",
```
eventualbuddha: Why pin it to 1.5.0?

src/stages/add-variable-declarations/index.js
```diff
@@ -9,7 +9,19 @@ export default class AddVariableDeclarationsStage {
...
+      if (err.loc) {
+        let {line, column} = err.loc;
+        let lines = content.split('\n');
+        let l = '-'.repeat(column) + '^';
+        lines.splice(line, 0, l, '').join('\n');
+        content = lines.join('\n');
+      }
+      err.message += '\n' + content;
+      throw(err);
+    }
```
eventualbuddha: This is great! Could this be generalized to all the post-main stages?

alangpierce:
Any updates on this? I was in the middle of writing something similar, but then noticed this open PR.

Looks like if we wanted it to work on all stages, we could probably just move the try/catch to be around `stage.run` in `src/index.js`? Although maybe that could cause excessive text to be added to exceptions from `NormalizeStage` or `MainStage` if they have `loc`? Another thing we could do to be more specific is use `instanceof SyntaxError`, I think.

Still, the `addVariableDeclarations` stage is by far the most useful. Almost all of the decaffeinate errors that I've seen are in that stage (since it's the first after `MainStage`), and showing the underlying bad code makes it way easier to figure out what's going wrong.

marcmenn:
I'm not currently working on this issue. As you pointed out, the `addVariableDeclarations` stage was my focus. I'd suggest merging this PR for now and leave the reuse in other stages for later. I'm not to familiar with the generic stage-handling code…

eventualbuddha:
Yep, I'll improve it myself if it seems warranted. Thanks!

alangpierce:
@marcmenn ok, thanks for the response. I think we don't want the package.json change, and I think I may be able to just move the code to the more general place like I said, so I'll see if I can send out a new PR that replaces this one.

alangpierce:
@eventualbuddha FYI I put up my approach at https://github.com/decaffeinate/decaffeinate/pull/272 , which replaces this PR. I think it should work, but also feel free to do your own thing if you think there's a better way to accomplish this. I do think it would be really great to have something like this as part of the official repl, since it's much more useful than a generic parser error.

eventualbuddha:
Superseded by #272. Sorry for the lag in getting this working.

Issue #220, opened 2016-05-15
error converting multi line single/double quoted strings. 


iainb:
example code that causes decaffeinate to crash:
```coffeescript
x = "double quoted
     multi line string"
```
becomes the following when converted to js
```js
var x;

x = "doube quoted multi line string";
```

decaffeinate leaves the new line in place and produces invalid code, from memory (when debugging) something similar to this:
```es6
let x = "double quoted
    multi line string"
```
possibly related issue in coffee-lex: https://github.com/decaffeinate/coffee-lex/issues/6

alangpierce:
Looks like this was fixed by https://github.com/decaffeinate/decaffeinate/commit/981c44ee6311e83e906e2512ec399d4cc2325274 , so should be safe to close.

PR #224, opened 2016-05-15
WIP - Improve multi line string support

iainb:
This is a work in progress, looking for a bit of feedback (it has broken tests, though they're of interest).

I started working on a fix for  #220 and discovered that there are some edge cases where decaffeinate does not convert triple quoted strings into template literals correctly or double quoted multi line strings that contain string interpolation (this is in addition to #220).

This PR starts to fix the problem (though its a little rough around the edges), it tries to convert strings correctly but does so at the expense of losing some of the original formatting.

An example 
```coffeescript
() ->
    return """multi line
              triple\nquoted
              string"""
```
Expected javascript
```javascript
(function() {
  return "multi line\ntriple\nquoted\nstring";
});
```
decaffeinate 2.10.0
```es6
() =>
    `multi line
              triple\nquoted
              string`
;
```
decaffeinate on this branch
```
() => `multi line\ntriple\nquoted\nstring`;
```

Based on the above, a few questions.
1. Before going on to fixing the newly broken tests, Is converting the multi line strings in this way sensible?
2. There is some nasty code in the StringPatcher (see patch method) any thoughts on how to make this better?

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "babel-preset-es2015-rollup": "^1.1.1",
     "babel-register": "^6.7.2",
     "browserify": "^13.0.0",
+    "coffee-script": "^1.10.0",
     "eslint": "^2.10.1",
     "eslint-plugin-babel": "^3.2.0",
     "eslint-plugin-decaffeinate": "file:eslint-rules",
--- a/src/stages/main/index.js
+++ b/src/stages/main/index.js
@@ -49,6 +49,7 @@ import SoakedFunctionApplicationPatcher from './patchers/SoakedFunctionApplicati
 import SoakedMemberAccessOpPatcher from './patchers/SoakedMemberAccessOpPatcher.js';
 import SpreadPatcher from './patchers/SpreadPatcher.js';
 import SuperPatcher from './patchers/SuperPatcher.js';
+import StringPatcher from './patchers/StringPatcher.js';
 import SwitchCasePatcher from './patchers/SwitchCasePatcher.js';
 import SwitchPatcher from './patchers/SwitchPatcher.js';
 import TemplateLiteralPatcher from './patchers/TemplateLiteralPatcher.js';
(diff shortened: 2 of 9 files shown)


madskonradsen:
Out of curiosity, shouldn't it keep the linebreaks, rather than inserting \n?

iainb:
You could keep the line breaks and remove the indentation. That's the alternative.

Here is a link to the [different decaffeinate outputs run through babel](https://babeljs.io/repl/#?evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&code=()%20%3D%3E%0A%20%20%20%20%60multi%20line%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20triple%5Cnquoted%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20string%60%0A%3B%0A%0A()%20%3D%3E%20%60multi%20line%5Cntriple%5Cnquoted%5Cnstring%60%3B%0A%0A()%20%3D%3E%0A%20%20%20%20%60multi%20line%0Atriple%5Cnquoted%0Astring%60%0A%3B)

Unfortunately neither of these styles are particularly nice compared with the CoffeeScript version.

madskonradsen:
You're right, the coffeescript version is nicer! It's just that I have a lot of 20-30+ lines-coffeescript-strings(Angular templates) and it would mess stuff like that up if they got compiled into one line with \n's :P In that case it would be preferable to keep the line breaks and just remove the indentation.

iainb:
Pushed some more changes, though its still quite a long way from being ready to merge. I've switched to creating template literals with actual line breaks in instead of literal '\n' (as per @madskonradsen's feedback).

Trying to handle strings properly is a quick way to lose your mind, as there are quite a large number of permutations. Writing tests for each one and making sure they're right is too  time consuming. Due to this I've added some integration style tests using the Coffee Script compiler, see: ebc6dab which is proving quite useful at shaking out bugs.

PR #239, opened 2016-06-05
Improve support for multi line strings

iainb:
Correctly handling multi line strings, here strings and both types that contain interpolation has a surprising number of edge cases. Because of this I've only added support for the most basic type of multi line string in this PR. It does not include support for here strings or any multi line string that contains interpolation.

Probably the most useful part of this PR is the inclusion of what I've called a 'string integration test' that compares the strings after being run through both decaffeinate + babel and coffee-script by evaluating the resulting es5 code and comparing the output. This technique was invaluable (though the tests are a little slow) as manually constructing test cases becomes very difficult. It also found many edge cases I hadn't considered, which is why the patchMultilineString function is so complicated.

If you add in here string quotes and interpolation test strings into the string_integration_test you'll find there are quite a few failure scenarios which decaffeinate currently doesn't handle correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
   ],
   "devDependencies": {
     "babel": "^6.5.2",
+    "babel-core": "^6.9.1",
     "babel-eslint": "^6.0.4",
     "babel-plugin-syntax-class-properties": "^6.5.0",
     "babel-plugin-syntax-flow": "^6.5.0",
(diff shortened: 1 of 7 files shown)


iainb:
re: tests - looks like the integration test doesn't pass/work correctly on anything lower than node 6.0

madskonradsen:
I love your "string integration test". It looks very effective even though it takes a bit of extra time to evaluate!

eventualbuddha:
This is great! I'll dig in soon.

iainb:
Tests fixed, was missing a package (babel-core) and increased the timeout for the integration tests.

eventualbuddha:
Merged manually in 981c44ee6311e83e906e2512ec399d4cc2325274. Thanks!

PR #249, opened 2016-06-13
Fix single parameter function edge cases

bhvaleri:


src/stages/main/patchers/BoundFunctionPatcher.js
```diff
@@ -65,6 +66,17 @@ export default class BoundFunctionPatcher extends FunctionPatcher {
...
 
+  parameterListNeedsParentheses(): boolean {
+    let parameters = this.parameters;
+
+    if (parameters.length !== 1) {
+      return true;
+    }
+
+    let [ param ] = parameters;
+    return (param instanceof IdentifierPatcher);
```
eventualbuddha: This seems like the reverse of what we actually want. I think this should be `!(param instanceof IdentifierPatcher)` and the usage on L56 should be `} else if (!this.parameterListNeedsParentheses()) {`.

bhvaleri: yep thanks, fixed

eventualbuddha:
Fixed in v2.13.1 ([repl](http://decaffeinate.github.io/decaffeinate/repl/#?evaluate=false&code=(args...)%20%3D%3E)).

Issue #271, opened 2016-07-01
"return if" can cause an error


alangpierce:
Here's a reduced case that fails:
```
f = ->
  return if true then null
```
http://decaffeinate.github.io/decaffeinate/repl/#?evaluate=true&code=f%20%3D%20-%3E%0A%20%20return%20if%20true%20then%20null

With some improved logging I have, it shows this error in an intermediate JS output:
```
an intermediate stage failed to parse: Unexpected token (2:9)
  1 | f = function() {
> 2 |   return if (true) { null; };
    |          ^
  3 | };
  4 | 
```

So it looks like `if` is being written as a statement when it should be converted to an expression.

Issue #270, opened 2016-07-01
Parentheses in string literals can cause errors


alangpierce:
Here's a small test case demonstrating this:
```
a(')').b
```
http://decaffeinate.github.io/decaffeinate/repl/#?evaluate=true&code=a(')').b

It gives the error "cannot find '.' in member access".

That was the problem I specifically ran into, but an even simpler case that fails is just `a(')')`, which produces this incorrect intermediate JS (using some improved error logging I have set up):
```
an intermediate stage failed to parse: Unexpected token (1:5)
> 1 | a(')';)
    |      ^
  2 | 
```

alangpierce:
Looks like this was fixed by the change to trimNonMatchingParentheses.js in 
https://github.com/decaffeinate/decaffeinate-parser/commit/2684933f9ab44ca725542dc1957e30721c564ab1 , so just waiting on a release of decaffeinate-parser and then decaffeinate.

eventualbuddha:
[Fixed in v2.17.2](http://decaffeinate-project.org/repl/#?evaluate=false&code=if%20a%20%26%26%20%28b%3D%3D'%29'%29%0A%20%20x).

Issue #269, opened 2016-07-01
for loops over implicit function calls crash


alangpierce:
Here's a reduced case:
```
for a in b c
  d()
```
http://decaffeinate.github.io/decaffeinate/repl/#?evaluate=true&code=for%20a%20in%20b%20c%0A%20%20d()

It gets an "unexpected token" error due to generating invalid JS code. With some improved error handling set up, I can get the incorrect JS output:
```
an intermediate stage failed to parse: Unexpected token (2:39)
  1 | iterable = b(c)
> 2 | for (i = 0; i < iterable.length; i++) {)
    |                                        ^
  3 |   a = iterable[i];
  4 |   d();
```

So it looks like it's putting a close-paren at the end of the for loop line, and it's causing a syntax error.

alangpierce:
Ok, I've tracked down the problem and I think I have a good fix for it. Here's what's going wrong:
* First, the function call `b c` is patched to become `b(c)`, the close-paren is inserted using magic-string's `insertLeft`, which seems correct to me (although the decaffeinate code path is `FunctionApplicationPatcher.patchAsExpression` calling `NodePatcher.insert` calling `NodePatcher.insertRight` calling `MagicString.insertLeft`, so probably that could be cleaned up).
* The code for the function is grabbed using magic-string's `slice` method. This respects the `insertLeft` so the close-paren is included when defining the `iterable` variable.
* The `for` loop header is rewritten using magic-string's `overwrite`. However, unlike `slice` and `move`, `overwrite` does NOT consider the `insertLeft`ed close-paren to be part of the range, so the close-paren insert stays and shows up in the output, so the resulting JS has that extra paren that breaks things. It might be controversial, but my impression is that this is just a bug in magic-string, and that all magic-string functions that take a range should have the same behavior as far as what's included in that range.

I have a PR for magic-string that fixes this case and doesn't cause any decaffeinate test failures, so I'll send it out soon.

PR #272, opened 2016-07-01
Improve error reporting when there is a parser error

alangpierce:
This is still sort of work-in-progress in that it copies `lineColumnMapper` from
`decaffeinate-parser`, so let me know if you think it would be better to change
`decaffeinate-parser` first to expose that function (or some other approach).

This is originally based on https://github.com/decaffeinate/decaffeinate/pull/222 ,
but I ended up changing the approach pretty significantly.

Since the `PatchError` class already contains logic to pretty-print an error in
a particular place in a file, with necessary context, I decided to reuse that.
Unfortunately, it relied on having a `ParseContext` from `decaffeinate-parser`,
even though none of the code is CoffeeScript-specific. The main purpose of the
`ParseContext` over just the source code is for the `lineMap`, so for now I
refactored `PatchError` to just take the source code and create its own
`lineMap` by copying the function from `decaffeinate-parser`. But maybe a better
solution is to just break the abstraction a little and expose the
`lineColumnMapper` function from `decaffeinate-parser`? Or maybe there's some
other clever way to reuse this code that I haven't thought of. Since it's just
20 lines, it seems like it's not worth worrying about that much, I think.

This change has been really useful for me to track down the underlying cause of
errors in files I'm trying to decaffeinate, and makes it much easier to change
files to be decaffeinate-friendly. See #269, #270, and #271 for examples of the
output produced after this change.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,6 +3,7 @@ import SemicolonsStage from './stages/semicolons/index.js';
 import EsnextStage from './stages/esnext/index.js';
 import MainStage from './stages/main/index.js';
 import NormalizeStage from './stages/normalize/index.js';
+import PatchError from './utils/PatchError.js';
 
 export { default as run } from './cli';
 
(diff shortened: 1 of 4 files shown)


eventualbuddha:
Thanks! I extracted a version of `lineColumnMapper` into the [lines-and-columns](https://www.npmjs.com/package/lines-and-columns) package a while ago but didn't get around to actually using it in both until now.

alangpierce:
Cool, that works out pretty well, then. Thanks!

Issue #278, opened 2016-07-04
Show the full context when reporting errors in the repl


alangpierce:
I recently made it so errors like "Unexpected token" show the full context in the CLI, but it turns out the repl just takes `error.message` rather than pretty-printing the error. It would be especially nice to have the full context in the repl to make it easy to quickly go through a file and make it decaffeinate-friendly.

I have two PRs for this coming up soon, just filing this issue for tracking. Here's a sneak peek:
http://www.alangpierce.com/decaffeinate/repl/#?evaluate=true&code=(a)%20-%3E%0A%20%20(a%20not%20in%20b.map(a%2C%20(e)%20-%3E%20%0A%20%20%20%20e)%0A%20%20)

niieani:
Fantastic job!

eventualbuddha:
Nice! Looking forward to the PRs!

eventualbuddha:
… which I see you already submitted. Will look soon!

alangpierce:
Fixed with https://github.com/decaffeinate/decaffeinate/commit/d77ebba2e88d594eed00516281fe33980b660e84 and https://github.com/decaffeinate/decaffeinate/commit/e10504110a5f16111a1851127545338a151ba9b9.

PR #279, opened 2016-07-04
Expose the PatchError class as a decaffeinate export

alangpierce:
Step 1 of 2 for #278.

This is useful so that the repl can call `PatchError.prettyPrint` on errors thrown
by `convert`.

This is the best I could come up with to both import `PatchError` for use in
`index.js` and also export it from `index.js`. If there's a way to do it that doesn't
involve typing `./utils/PatchError.js` twice, I'm happy to go with that.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -5,6 +5,7 @@ import MainStage from './stages/main/index.js';
 import NormalizeStage from './stages/normalize/index.js';
 import PatchError from './utils/PatchError.js';
 
+export { default as PatchError } from './utils/PatchError.js';
 export { default as run } from './cli';
 
 type Options = {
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -5,6 +5,7 @@ import MainStage from './stages/main/index.js';
 import NormalizeStage from './stages/normalize/index.js';
 import PatchError from './utils/PatchError.js';
 
+export { default as PatchError } from './utils/PatchError.js';
```
eventualbuddha: This can just be done as:

```js
export { PatchError };
```

alangpierce: Ah, thanks, I guess I was missing the braces when I tried that kind of thing.

PR #283, opened 2016-07-06
Fix PatchError.prettyPrint when the file doesn't end in a newline

alangpierce:
The code was using `isNaN` to determine if the line/column pair was valid, but
`isNaN(null)` is `false` in JS, so this was always returning false. This caused
the repl to sometimes not display the last line in the error, e.g.:
http://decaffeinate.github.io/decaffeinate/repl/#?evaluate=true&code=a%3F.b

### DIFF
--- a/src/utils/PatchError.js
+++ b/src/utils/PatchError.js
@@ -52,8 +52,8 @@ export default class PatchError extends Error {
     for (let line = displayStartLine; line <= displayEndLine; line++) {
       let startOfLine = lineMap.indexForLocation({ line, column: 0 });
       let endOfLine = lineMap.indexForLocation({ line: line + 1, column: 0 });
-      if (endOfLine == null) {
-        if (startOfLine == null) {
+      if (endOfLine === null) {
+        if (startOfLine === null) {
           break;
         } else {
           endOfLine = source.length;
--- a/src/utils/PatchError.js
+++ b/src/utils/PatchError.js
@@ -52,8 +52,8 @@ export default class PatchError extends Error {
     for (let line = displayStartLine; line <= displayEndLine; line++) {
       let startOfLine = lineMap.indexForLocation({ line, column: 0 });
       let endOfLine = lineMap.indexForLocation({ line: line + 1, column: 0 });
-      if (isNaN(endOfLine)) {
-        if (isNaN(startOfLine)) {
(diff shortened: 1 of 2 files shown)


src/utils/PatchError.js
```diff
@@ -52,8 +52,8 @@ export default class PatchError extends Error {
     for (let line = displayStartLine; line <= displayEndLine; line++) {
       let startOfLine = lineMap.indexForLocation({ line, column: 0 });
       let endOfLine = lineMap.indexForLocation({ line: line + 1, column: 0 });
-      if (isNaN(endOfLine)) {
-        if (isNaN(startOfLine)) {
+      if (endOfLine == null) {
```
eventualbuddha: Should use `===`.

alangpierce: Ok, I just updated it. FYI I saw that the flow type for `indexForLocation` is `?number`, which means "undefined or null or a number" (I think), so I've found that a `== null` check (i.e. checking for `null` and `undefined` at the same time) tends to make flow happier. Not sure if that really matters in this case, though.

eventualbuddha: The docs for nullable types don't say anything about `undefined`, at least that I can find: https://flowtype.org/docs/five-simple-examples.html#2-adding-type-annotations.

eventualbuddha: In any case, I haven't actually enabled flow checking for this project yet. Only coffee-lex has it of the three.

alangpierce: :thumbsup: yeah, I noticed that flow checking wasn't enabled, just calling that out as maybe something to think about. In a flow project I worked on recently, I ended up deciding to always use `== null` and `!= null` when working with `?` types.

I'm not sure where it would be in the docs (if anywhere), but here's an example:
https://tryflow.org/?code=bGV0IGZvbzogP251bWJlciA9IG51bGw7CmxldCBiYXI6IG51bWJlciA9IGZvbyA9PT0gbnVsbCA/IDAgOiBmb287CmxldCBiYXo6IG51bWJlciA9IGZvbyA9PSBudWxsID8gMCA6IGZvbzs=

The second line doesn't typecheck because flow thinks that a `?number` could still be `undefined` after a `=== null` check.

PR #290, opened 2016-07-08
fix: always use an expression as the argument to a return statement

### LINKED ISSUE #271: "return if" can cause an error

alangpierce:
This comes up, for example, when doing `return if ...`.

Closes #271.

### DIFF
--- a/src/stages/main/patchers/ReturnPatcher.js
+++ b/src/stages/main/patchers/ReturnPatcher.js
@@ -11,6 +11,9 @@ export default class ReturnPatcher extends NodePatcher {
 
   initialize() {
     this.setExplicitlyReturns();
+    if (this.expression !== null) {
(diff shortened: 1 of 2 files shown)


eventualbuddha:
Good catch!

Issue #285, opened 2016-07-07
Provide sample CoffeeScript in repl when no query string exists


eventualbuddha:
Probably code taken from CoffeeScript.org

madskonradsen:
I believe this have been fixed :)

alangpierce:
Looks like it shows the sample code for a split second when loading existing code in the repl, which would be nice to fix.

![visual_issue](https://cloud.githubusercontent.com/assets/211605/16696322/3d0a9368-44f9-11e6-8cf7-c21130dc1449.gif)

PR #292, opened 2016-07-08
fix: only add demo-code in case of no query-string or localstate is present

madskonradsen:
Fixes the issue mentioned in #285

The demo-code is kept in a hidden text-field until it is decided whether or not it should be used as it seemed like the most sane solution.


madskonradsen:
Squashing commits is apparently the hardest thing in the world...

eventualbuddha:
Thanks!

PR #300, opened 2016-07-10
add a warning about how to test against an unreleased decaffeinate-parser

alangpierce:
I recently tried using `npm link` to test decaffeinate against a local
decaffeinate-parser, and ran into really confusing behavior that took quite a
while to figure out. Rather than failing in an obvious way, it would put extra
parens in confusing places and other random-seeming problems like that. The
problem ended up being that decaffeinate and decaffeinate-parser were using
different instances of coffee-lex, so comparisons between token types, like
`token.type === IDENTIFIER` would always be false (since it's an identity
equality check on singletons).

To make this case more visible, this commit adds an up-front check with an
explanation of the issue and a suggestion for how to properly set things up.

To test this, I verified that no warning is shown under a normal configuration,
and that if you use `npm link` to hook up decaffeinate-parser, all tests fail with
this error. (I had to mess with some recently-introduced flow settings to get
flow to pass after running npm link.)

Note that this could come up in an actual build, I think, since there's no
guarantee that decaffeinate and decaffeinate-parser pull in the same version.
Maybe that would be fixed using a peer dependency or something like that?

Also, if there's a better workflow for testing unreleased versions, I would be
happy to hear it.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -4,6 +4,7 @@ import EsnextStage from './stages/esnext/index.js';
 import MainStage from './stages/main/index.js';
 import NormalizeStage from './stages/normalize/index.js';
 import PatchError from './utils/PatchError.js';
+import verifyCoffeeLexLoadedOnce from './utils/verifyCoffeeLexLoadedOnce.js';
 
 export { default as run } from './cli';
 export { PatchError };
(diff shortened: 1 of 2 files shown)


eventualbuddha:
I like the warning, but I think it would be just as effective to call once at the beginning of the tests. The instructions also just list out a set of steps that could easily be automated. Maybe have a dev script that clones all the decaffeinate repos and links them appropriately?

alangpierce:
I think I'll just close this, at least for now, since I'm not super-motivated to get it through, and it may not even come up for anyone else. Another approach rather than a runtime check is to just mention it in the docs somewhere (e.g. maybe the contribution guide could have a part with tips for working across multiple repos). I also don't see any obvious way to run it before any test without having to change every test file, although I may be overlooking something.

PR #302, opened 2016-07-11
fix: add implicit function call parens in the normalize step

### LINKED ISSUE #269: for loops over implicit function calls crash

### LINKED ISSUE #301: Method closing curly brace sometimes placed too late
Here's an example that's as reduced as I can get it:
```
A = {
  b: ->
    return c d,
      e,
        f
}
G
```

It gives this error:
```
AddVariableDeclarationsStage failed to parse: Unexpected token (7:0)
  5 |         f)
  6 | }
> 7 | G;
    | ^
  8 |   };;
```

alangpierce:
The code in #301 is actually parsed incorrectly by the CoffeeScript parser (even
before decaffeinate-parser does anything interesting). It says that the function
body ends at the end of the file, so decaffeinate was inserting a
close-curly-brace at the end of the file, which was incorrect. However, the
function application was correctly parsed, and explicitly putting parens for the
function application causes the CoffeeScript parser to work again, so we can
work around this issue by inserting parens for all implicit functions before the
MainStage.

I think this will also fix a bunch of other problems caused by implicit function
calls. For example, it also happens to fix #269. In the codebase that I'm trying
to decaffeinate, it fixes 49 out of the 104 files with decaffeinate failures
that I hadn't categorized.

I added the heuristic that we put the close-paren on the next line if the last
arg is a multi-line expression. This fixed the formatting in almost every test,
except the change introduced two test issues:
* Object literal formatting sometimes had an excessive newline in non-implicit
  function calls
  ( http://decaffeinate-project.org/repl/#?evaluate=true&code=a(b%2C%0A%20%20c%3A%20d%0A%20%20e%3A%20f%0A%29 )
  so adding parens exposed the issue. It's pretty minor, though. I think.
* In another test, the added parens interfered with some other operations in the
  normalize step in a way that will be fixed by https://github.com/Rich-Harris/magic-string/pull/89

Closes #301.
Closes #269.

### DIFF
--- a/test/function_test.js
+++ b/test/function_test.js
@@ -156,4 +156,24 @@ describe('functions', () => {
   it('keeps function with a single assignment as a parameter in braces', () => {
     check(`(args=false) =>`, `(args=false) => {};`);
   });
+
+  it('places the function end in the right place when ending in an implicit function call', () =>
+    check(`
+      A = {
+        b: ->
+          return c d,
+            e,
+              f
+      }
+      G
+    `, `
+      const A = {
+        b() {
+          return c(d,
+            e,
+              f);
+        }
+      };
+      G;
+    `));
 });
--- a/src/patchers/NodePatcher.js
+++ b/src/patchers/NodePatcher.js
@@ -656,7 +656,7 @@ export default class NodePatcher {
    * Determines whether this patcher's node spanned multiple lines.
    */
   isMultiline(): boolean {
-    return /[\r\n]/.test(this.getOriginalSource());
+    return !this.node.virtual && /[\r\n]/.test(this.getOriginalSource());
   }
 
   /**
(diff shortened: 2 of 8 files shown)


test/function_test.js
```diff
@@ -156,4 +156,24 @@ describe('functions', () => {
...
+            e,
+              f
+      }
+      G
+    `, `
+      const A = {
+        b() {
+          return c(d,
+            e,
+              f);
```
eventualbuddha: Ideally I think this paren would be on the next line. Pretty minor though, can be done later.

Issue #268, opened 2016-07-01
Add support for expansion nodes


alangpierce:
Simple example:

```
[a, ...] = b
```

https://decaffeinate.github.io/decaffeinate/repl/#?evaluate=true&code=%5Ba%2C%20...%5D%20%3D%20b

This currently gives the exception "Error: cannot traverse unknown node type: Expansion" along with a giant stack trace. Assuming the implementation isn't a quick fix, it would be nice to give a more friendly error message pointing at the offending token in the code and saying that the feature isn't yet implemented.

alangpierce:
The remaining issue here is handling nested expansions. For example:

```coffee
[[a, ..., b]] = arr
```

[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=%5B%5Ba%2C%20...%2C%20b%5D%5D%20%3D%20arr)

which gives this error:

```
expansions (e.g. `[a, ..., b] = c`) are not supported yet in all cases, see https://github.com/decaffeinate/decaffeinate/issues/268
> 1 | [[a, ..., b]] = arr
    |      ^^^
```

PR #309, opened 2016-07-13
feat: handle expansions in array assignment

alangpierce:
This is a partial implementation of expansions, which handles them for array
assignment and destructuring, but not for function parameters or nested within
more complicated assignments. For those unsupported cases, I throw an error
linking to #268 when the expansion node is traversed.

Currently the only special case is `[a, b, ...]`. Every other case just turns
into a series of assignments that do array lookups.

### DIFF
--- a/src/stages/main/patchers/AssignOpPatcher.js
+++ b/src/stages/main/patchers/AssignOpPatcher.js
@@ -1,3 +1,5 @@
+import ArrayInitialiserPatcher from './ArrayInitialiserPatcher.js';
+import ExpansionPatcher from './ExpansionPatcher.js';
 import NodePatcher from './../../../patchers/NodePatcher.js';
 import type { Node, ParseContext, Editor } from './../../../patchers/types.js';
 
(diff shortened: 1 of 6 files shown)


src/stages/main/patchers/AssignOpPatcher.js
```diff
@@ -17,14 +19,123 @@ export default class AssignOpPatcher extends NodePatcher {
...
+
+    assignees.forEach((assignee, i) => {
+      // Patch everything but the expansion node, since expansion nodes expect
+      // to not be patched.
+      if (i !== expansionIndex) {
+        assignee.patch();
+      }
+    });
+    this.expression.patch();
+    let expressionCode = this.slice(this.expression.contentStart, this.expression.contentEnd);
```
eventualbuddha: This can be done using `this.expression.getOriginalCode()`. It's not exactly the same in all cases since `this.slice()` uses the patched string. In this case `this.expression` shouldn't have been patched yet so it should be the same.

eventualbuddha:
Great, thanks!

Issue #156, opened 2016-03-25
Add support for `for` loops used as expressions


eventualbuddha:
Including implicit returns, e.g.

```coffee
a(b for b in c)
->
  for a in b
    a
```

d4rky-pl:
Is there any workaround?

kazazor:
Hi @eventualbuddha - I can see that you've already developed this issue, but I still getting this error when trying to use it:
```'for' loops used as expressions are not yet supported (https://github.com/decaffeinate/decaffeinate/issues/156)```

It is truly on master and being used by this [repl](http://decaffeinate.github.io/decaffeinate/repl/)?

**Comment by sethcall** [comment]
I was also going to ask, say I have something like this:
```
Junk = {
  c: () ->
    d

  d: () ->

    for a in b
      @c()

}
```

What do I do to re-write it such that I can workaround this decaffeinate limitation?  It's not clear to me that there is a way out on this. Thanks for any help.

eventualbuddha:
The easiest way to work around it is just to add a `return` after the loop at the end of the function.

sethcall:
That works!

alangpierce:
Hey @eventualbuddha, I'm thinking of tackling this one pretty soon (or at least part of it). I wonder if you have thoughts on any challenges or edge cases. I saw that the normalize step gets rid of the post-for form, so that should make things simpler.

My current thought is to rewrite `for...in` like this:
`for a in b then c` -> `b.map(a => c)`

And when there's a filter, something like this:
`for a in b when c then d` -> `b.filter(a => c).map(a => d)`

For `for...of` I was thinking of doing a similar thing using `Object.keys` and `Object.entries`, but I guess that only goes over own properties and not properties in the prototype chain. Maybe we could emit a helper that generates the right properties and just filter/map on that? (It looks like `for...in` is much more common in the codebase I'm porting, though, so I may just focus on that.)

I guess another concern is that if the body has a `break`, `continue`, or `return`, then it wouldn't work inside the lambda. That seems kind of rare, though (especially for expression for loops), so maybe as a first pass I could just fail for that case.

Does that seem reasonable? I saw in some of the test cases in `for_test.js` that you have expected code using an IIFE with a regular for loop, so I wonder if there's more that I haven't thought of. I guess technically `map` wouldn't work if the target is an array-like object that exposes `length` and indexing at int indices, but doesn't have a `map` method. Do you think that's a case we care about?

[...]

PR #321, opened 2016-07-16
feat: handle some "for in" loop expressions using filter and map

alangpierce:
Partial fix for #156.

This commit handles "for in" using the strategy of transforming it to
`a.filter(...).map(...)`. This works in the most common cases, but there are some
edge cases that it doesn't handle (where I throw an error):
* Loop expressions using a `by` clause.
* Loop expressions where the body cannot be patched as an expression. In
  particular, I don't think there's a great way to handle `break`, `continue`,
  and `return` within loop expressions using `map` and a lambda.
* Expressions that use a filter and also assign the index to a variable. In this
  case, the code would be incorrect, because the filter step changes the size of
  the array and thus the indexes.

This technically could change behavior when looping over array-like objects,
like `{length: 2, "0": "hello", "1", "world"}`. But I think that's obscure
enough that it's unlikely to come up in practice, and if it does, it's better to
generate the nice code by default and the 100% correct code as an option.

I also added some skipped tests for cases that this commit doesn't support.

Out of the 180 files in the codebase I'm working with that were failing due to
for loop expressions, this commit fixes 159 of them. Of the remaining 21 of
them, 11 of them fail due to a non-expression body (about half of which had
`continue` statements), 4 of them fail due to a "for of" expression, 4 of them
fail because they use a "by" clause, and 2 of them fail because they use
"for own".

### DIFF
--- a/src/stages/main/patchers/ForInPatcher.js
+++ b/src/stages/main/patchers/ForInPatcher.js
@@ -16,6 +16,67 @@ export default class ForInPatcher extends ForPatcher {
     }
   }
 
+  patchAsExpression() {
+    if (this.step !== null) {
+      throw this.error(
+        `'for in' loop expressions with a "by" clause are not supported yet ` +
+        `(https://github.com/decaffeinate/decaffeinate/issues/156)`
+      );
+    }
+    if (!this.body.canPatchAsExpression()) {
+      throw this.error(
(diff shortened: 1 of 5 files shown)


eventualbuddha:
Thanks!

niieani:
awesome

PR #324, opened 2016-07-18
feat: implement soaked member access, dynamic member access, and function application

### LINKED ISSUE #176: no patcher available for node type: SoakedFunctionApplication
I've tried decaffeinating this file: https://github.com/rethinkdb/rethinkdb/blob/next/drivers/javascript/cursor.coffee

But I'm getting the error:

```
cursor.coffee: no patcher available for node type: SoakedFunctionApplication (props: function, arguments)
  223 |             if err?
  224 |                 if err.message is 'No more rows in the cursor.'
> 225 |                     onFinished?()
      |                     ^^^^^^^^^^^^^
  226 |                 else
  227 |                     cb(err)
```

### LINKED ISSUE #216: no patcher available for node type: SoakedDynamicMemberAccessOp
```
$ echo 'myObject?[myKey]' | decaffeinate
stdin: no patcher available for node type: SoakedDynamicMemberAccessOp (props: expression, indexingExpr)
> 1 | myObject?[myKey]
    | ^^^^^^^^^^^^^^^^
  2 |
```

alangpierce:
This is a first pass that I think should correctly handle nearly every case for
all soak operations. For now, the generated JavaScript always uses a helper
function, but in the future it probably makes sense to do something
simpler/shorter when we can get away with it.

Some details:
* In order to handle chained soaked operations, like `a?.b.c = d()`, I walk up
  the AST and stop when I get to the last soak-friendly node. That node's start
  and end define the "soak range" for the `__guard__` function.
* To handle nested soaks correctly, we also stop traversing up if we hit another
  soak operation.
* Since function application soaks have a different condition (they only pass if
  left side is really a function), I had to make a separate `__guardFunc__`
  helper for that case.
* I named the functions `__guard__` and `__guardFunc__` for now, since
  `registerHelper` doesn't ensure that the name is unique. Maybe in the future
  they could just be `guard` and `guardFunc`, or maybe something shorter.
* I process the nodes from the outside in. This makes sure that the guard
  functions are written in the right order, which is necessary for the last
  test.
* CoffeeScript also handles prefix and postfix ++ and -- and the delete
  operation. The prefix operators don't work with our strategy of appending code
  to the right, but it can probably be made to work. For now, though, I just
  throw an error.

Before this change, out of 1330 files in the codebase I'm working in, 235 of
them failed due to this error, and this commit completely fixes 166 of those
files, and the rest now fail due to other errors.

Closes #176.
Closes #216.

### DIFF
--- a/test/soaked_test.js
+++ b/test/soaked_test.js
@@ -1,12 +1,16 @@
 import check from './support/check.js';
+import validate from './support/validate.js';
 
-describe.skip('soaked expressions', () => {
+describe('soaked expressions', () => {
   describe('function application', () => {
     it('works with a basic function', () => {
       check(`
         a?()
       `, `
-        if (typeof a === 'function') { a(); }
+        __guardFunc__(a, f => f());
+        function __guardFunc__(func, transform) {
+          return typeof func === 'function' ? transform(func) : undefined;
+        }
       `);
     });
 
@@ -14,8 +18,10 @@ describe.skip('soaked expressions', () => {
       check(`
         a()?()
       `, `
-        let fn;
-        if (typeof (fn = a()) === 'function') { fn(); }
+        __guardFunc__(a(), f => f());
+        function __guardFunc__(func, transform) {
+          return typeof func === 'function' ? transform(func) : undefined;
+        }
       `);
     });
 
@@ -23,8 +29,10 @@ describe.skip('soaked expressions', () => {
       check(`
         a(b()?())
       `, `
-        let fn;
-        a(typeof (fn = b()) === 'function' ? fn() : undefined);
+        a(__guardFunc__(b(), f => f()));
+        function __guardFunc__(func, transform) {
+          return typeof func === 'function' ? transform(func) : undefined;
+        }
       `);
     });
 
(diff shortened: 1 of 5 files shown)


test/soaked_test.js
```diff
@@ -1,143 +1,348 @@
...
+          return typeof func === 'function' ? transform(func) : undefined;
+        }
       `);
     });
+
+    it('evaluates soaked function calls', () => {
+      validate(`
+        f = -> 3
+        o = f?()
+      `, 3);
```
alangpierce: Note that I'm using the change to `validate` from #321 here. Without that change, the `3` argument is ignored (although it's still useful documentation, in my opinion), but once #321 lands, `validate` will also make sure that the result is `3` (not just that CoffeeScript and Decaffeinate produce the same result).

niieani:
Such fantastic work 👑 ! Thanks @alangpierce.

eventualbuddha:
Thanks for all the improvements, @alangpierce! I still would prefer turning these into `if` statements, but this is much better than failing.

alangpierce:
Thanks for merging! Agreed that the generated code isn't always ideal. I just filed #336 to use `if` or other approaches when possible.

PR #355, opened 2016-07-30
feat: add a `runToStage` option so decaffeinate can return intermediate results

alangpierce:
This is the decaffeinate side of a change to the repl that makes it possible to
run some but not all stages of decaffeinate and inspect the intermediate
results. In addition to supporting the normal stages, I also allow inspecting
the results produced by the CoffeeScript parser, coffee-lex, and
decaffeinate-parser. In all three of these cases, there wasn't (to my knowledge)
an existing nice way of printing out these formats, so I wrote pretty-printers
for each of them, which use a consistent format for naming positions in code.

There are certainly some similarities between the pretty-printers, but there are
enough differences that it seems best to keep the code mostly separate.

The main use case here is to make it easier to figure out what went wrong when
decaffeinate crashes, e.g. whether the problem happened due to a bug in the
parser or at the normalize step or in the main stage. It may also make it easier
for future contributors to learn the internals of decaffeinate.

Here's a live demo:
http://www.alangpierce.com/decaffeinate-project.org/repl/#?evaluate=false&stage=full&code=x%20%3D%20a%20%22%23%7Bb%20%2B%20c%7D%22

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,13 +3,18 @@ import SemicolonsStage from './stages/semicolons/index.js';
 import EsnextStage from './stages/esnext/index.js';
 import MainStage from './stages/main/index.js';
 import NormalizeStage from './stages/normalize/index.js';
+import formatCoffeeLexAst from './utils/formatCoffeeLexTokens.js';
+import formatCoffeeScriptAst from './utils/formatCoffeeScriptAst.js';
+import formatDecaffeinateParserAst from './utils/formatDecaffeinateParserAst.js';
+import parse from './utils/parse.js';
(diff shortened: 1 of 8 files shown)


eventualbuddha:
That demo is great! Definitely cool. I'll take a look more when I get a chance. Playing with the demo, I wish it acted like http://astexplorer.net/ (highlighting parts of source or AST & vice-versa). That can be something for another time, though.

alangpierce:
Yep, agreed that something more interactive would be nice, but the low-tech text-based solution is the easy first pass. Hopefully the common format for source code ranges would make that a little easier to implement.

PR #362, opened 2016-08-04
feat: implement "for own" in a statement context

### LINKED ISSUE #157: Add support for `for own` loops
```coffee
for own a of b
  a
```

alangpierce:
We can just use a JS for-of loop on `Object.keys`, which is defined to iterate
through the own properties in the same order that JS for-in would.

Closes #157.

### DIFF
--- a/test/for_test.js
+++ b/test/for_test.js
@@ -475,67 +475,58 @@ describe('for loops', () => {
     `);
   });
 
-  it.skip('handles `for own`', () => {
+  it('handles `for own`', () => {
     check(`
       for own key of list
         console.log key
     `, `
-      for (let key in list) {
-        if (Object.prototype.hasOwnProperty.call(list, key)) {
-          console.log(key);
-        }
+      for (let key of Object.keys(list)) {
+        console.log(key);
       }
     `);
   });
 
-  it.skip('handles `for own` with an unsafe-to-repeat iterable', () => {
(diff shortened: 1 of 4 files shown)


madskonradsen:
Love the prettier syntax <3

test/for_test.js
```diff
@@ -475,67 +475,58 @@ describe('for loops', () => {
...
     `);
   });
 
-  it.skip('handles single-line `for own`', () => {
+  it('handles single-line `for own`', () => {
     check(`
       for own a of b then a
     `, `
-      for (let a in b) { if (Object.prototype.hasOwnProperty.call(b, a)) { a; } }
+      for (let a of Object.keys(b)) { a; }
```
eventualbuddha: We should also test that unsafe-to-repeat lists work like this.

eventualbuddha: That is, when accessing the value at each key.

alangpierce: Good call, I thought that would just work like it works for normal `for of`, but it breaks here. I'll fix and add a test.

alangpierce:
Ok, I just updated the commit to handle non-repeatable targets where both the key and value are bound. I ended up just reworking the code to use the same code and approach as `ForInPatcher`, which also makes some other cases look nicer.

Another thought here is to loop over `Object.entries`, so the expression doesn't need to be repeated in the first place, but for now I just stuck with the existing approach.

eventualbuddha:
Thanks!

Issue #356, opened 2016-07-30
CoffeeScript parser gives incorrect location data in some cases


alangpierce:
I noticed in some cases that Decaffeinate would generate the close-paren for an implicit function call in the wrong place, and I tracked it down to a bug in the CoffeeScript parser.

Here's an example where Decaffeinate generates incorrect code:
```
a {
  b: ->
    return c d,
      if e
        f
}
g
```
[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&code=a%20%7B%0A%20%20b%3A%20-%3E%0A%20%20%20%20return%20c%20d%2C%0A%20%20%20%20%20%20if%20e%0A%20%20%20%20%20%20%20%20f%0A%7D%0Ag)

It treats `g` as being an argument to `a`, which is wrong. In other cases, this problem causes Decaffeinate to generate invalid JS code where the parens/braces are out of order.

The problem is that the CoffeeScript lexer generates the wrong location data for the OUTDENT tokens after the `f`, so Decaffeinate thinks the function call ends after the `g` and places the close-paren in the wrong place.

I tracked down the bug and fixed it and sent a PR here: https://github.com/jashkenas/coffeescript/pull/4291 . [Here's Decaffeinate working correctly with the fix applied in my repl.](http://www.alangpierce.com/decaffeinate-project.org/repl/#?evaluate=false&stage=full&code=a%20%7B%0A%20%20b%3A%20-%3E%0A%20%20%20%20return%20c%20d%2C%0A%20%20%20%20%20%20if%20e%0A%20%20%20%20%20%20%20%20f%0A%7D%0Ag)

However, even if my PR gets accepted, CoffeeScript hasn't had a release in 11 months, so it's unlikely we'd be able to use it in an official CoffeeScript release any time soon. Also, even working off of master is non-trivial because the AST format has had some changes since the 1.10.0 release.

So I think to fix this issue, we should fork CoffeeScript and work off of the fork. That would also give us the flexibility to make any other changes/fixes as necessary. @eventualbuddha, I'm sure you have thoughts on how to set this up, but here's the kind of thing I have in mind:
* Fork CoffeeScript into a repo in the decaffeinate org (probably decaffeinate/coffeescript).
* Create a branch off of the 1.10.0 tag that we'll use for decaffeinate-specific patches, and apply the patch from my PR.
[...]

PR #368, opened 2016-08-06
chore: add failing test for #356

alangpierce:
This will be fixed when updating to the first decaffeinate-coffeescript patch.


alangpierce:
Actually, I'll just abandon this and submit another PR that updates the dependency and adds the un-skipped test.

PR #376, opened 2016-08-09
fix: always spread `arguments` in `super` call with implicit args

### LINKED ISSUE #310: Wrong super keyword compilation.
decaffeinate version 2.17.1

```coffeescript
class C extends B
  constructor: (x) ->
    super

  fn: (x) ->
    super
```
```javascript
class C extends B {
  constructor(x) {
    super();
  }

  fn(x) {
    return super.fn();
  }
}
```
ES6 code should call super with `arguments` in this examples.

eventualbuddha:
Fixes #310

### DIFF
--- a/src/stages/normalize/patchers/FunctionApplicationPatcher.js
+++ b/src/stages/normalize/patchers/FunctionApplicationPatcher.js
@@ -23,6 +23,11 @@ export default class FunctionApplicationPatcher extends NodePatcher {
       return;
     }
 
+    if (this.isImplicitSuper()) {
+      this.insert(this.fn.outerEnd, '(arguments...)');
+      return;
+    }
+
     if (implicitCall) {
       let firstArg = args[0];
       let hasOneArg = args.length === 1;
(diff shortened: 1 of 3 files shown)


src/stages/normalize/patchers/FunctionApplicationPatcher.js
```diff
@@ -23,6 +23,12 @@ export default class FunctionApplicationPatcher extends NodePatcher {
       return;
     }
 
+    if (this.isImplicitSuper()) {
+      this.log(this.fn);
```
alangpierce: Maybe you left this in by mistake?

alangpierce:
LGTM. Too bad the AST format isn't more direct here.

src/stages/normalize/patchers/FunctionApplicationPatcher.js
eventualbuddha: Yep, thanks!

eventualbuddha:
> Too bad the AST format isn't more direct here.

Well, we do control that 😉

PR #381, opened 2016-08-13
fix: set decaffeinate-coffeescript as a normal dependency, fix test

alangpierce:
In #377, I made `index.js` depend on `decaffeinate-coffeescript` so that the
"run to stage" feature can directly access the CoffeeScript lexer, but I forgot
to move that dependency from `devDependencies` to `dependencies`. This caused
decaffeinate to break in a separate project where I had it installed through
package.json.

Also, the recent update to `decaffeinate-coffeescript` patch 2 causes the lexer
to give slightly different output, which broke a test, so update that.

While I'm at it, change the dependency to patch2.
(Although I think it gets picked up anyway due to [npm's semver rules](https://docs.npmjs.com/misc/semver#prerelease-tags).).


alangpierce:
Merging in now since this fixes a crash and a test break.

PR #392, opened 2016-08-19
feat: add support for all for loops used as expressions

### LINKED ISSUE #156: Add support for `for` loops used as expressions

alangpierce:
This changes `for` loops used as expressions to use a statement-style loop
within an IIFE, unless we already have a better way to transform the expression.

Conceptually, this was fairly easy since we already do this type of thing for
`while` loops, so it was mostly a matter of pulling that logic into a shared
`LoopPatcher` class. However, getting the formatting right in all cases ended up
being a huge pain, so I ended up reworking how a lot of that stuff works.

There are two main things that I introduced:
* All styles of loops now use common terminology for three indentation levels:
  the top-level loop where the `for` or `while` goes, the first level of the
  loop body, where we often generate an assignment statement, and the "main"
  loop body, which might be indented one more level to be wrapped in a generated
  `if` statement. Various code paths can now reference those logical indentation
  levels.
* To handle the three different cases of prepending a line to the front of a
  block, I added a new `insertLineBefore` method. In particular, this handles
  both the awkward inline-to-non-inline case, and also allows prepending lines
  that are at a lower indentation level than the block itself. A similar `insertLineAfter`
  method also ended up being useful.

One important property of the implementation here is that the only indentation
is in the loop body, and the indentation happens exactly once. I tried lots of
approaches involving indenting twice or indenting two different nodes, and all
of those seemed to confuse magic-string in various edge cases. Fewer operations
is better in this case, it seems.

Closes #156.

### DIFF
--- a/src/stages/main/patchers/BlockPatcher.js
+++ b/src/stages/main/patchers/BlockPatcher.js
@@ -150,6 +150,52 @@ export default class BlockPatcher extends NodePatcher {
   }
 
   /**
+   * Insert a statement before the current block. Since blocks can be patched in
+   * a number of ways, this needs to handle a few cases:
+   * - If it's completely inline, we don't deal with any indentation and just
+   *   put a semicolon-separated statement before the start.
+   * - If it's a normal non-inline block, we insert the statement beforehand
+   *   with the given indentation. However, `this.outerStart` is the first
+   *   non-whitespace character of the first line, so it's already indented, so
+   *   if we want to add a line with *less* indentation, it's a lot more tricky.
+   *   We handle this by walking backward to the previous newline and inserting
+   *   a new line from there. This allows the prepended line to have whatever
+   *   indentation level we want.
+   * - In some cases, such as nontrivial loop expressions with an inline body,
+   *   the source CoffeeScript is inline, but we want the result to be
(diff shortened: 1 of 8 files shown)


alangpierce:
Sorry this diff got a bit big and complicated. One thing to keep in mind is that most of `LoopPatcher` is moved directly from `WhilePatcher`.

alangpierce:
@eventualbuddha do you think you'll have time to get to this (and the other PR in this repo and the two PRs in the decaffeinate-parser repo)? I can just merge them if you're ok with that. They all have tests, so hopefully there isn't a lot that can go wrong :-) .

I can also try to find other contributors so changes still get a second pair of eyes. I may be able to convince someone at my company to help out, and there may be others watching the project who could help out.

kasbah:
I used your branch to translate [my browser extension](https://github.com/monostable/1clickBOM/tree/decaffeinate) with which I was getting `for..in` not supported errors. So thanks very much to you and the rest of the decaffeinate team!

eventualbuddha:
Sorry for the delay in looking at the PRs. Things have been busy lately and I've been sick (just a cold).

PR #443, opened 2016-10-02
fix: give a more clear error message on complex conditionals

alangpierce:
Previously, we would just skip patching IIFE-style conditionals, causing a
confusing error later, but now we give a helpful message pointing at a bug.

I also made it so we require both the consequent and the alternate (if present)
to be patchable as an expression before we say that the condition can be a
ternary; otherwise, it was hitting a more opaque error in some cases.


alangpierce:
Actually, nevermind, I think I can just implement this, so no need for the error. :-) PR coming soon.

PR #466, opened 2016-10-23
feat: allow code within class bodies

### LINKED ISSUE #397: Support arbitrary code in class bodies
There has already been some discussion about this in #293 and #123, but this bug is a bit more specific and direct, and I wanted to throw out an idea I have for how to implement it and see if anyone has feedback.

Currently, decaffeinate makes no specific attempt to handle code within class bodies, so you end up with a parser error on the JS side. Here's an example that fails:
```coffee
class Foo
  SOME_CONSTANT = 7
  computeThing: -> SOME_CONSTANT + 3
```
[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=class%20Foo%0A%20%20SOME_CONSTANT%20%3D%207%0A%20%20computeThing%3A%20-%3E%20SOME_CONSTANT%20%2B%203)

Unfortunately, JS fundamentally can't handle code within class bodies (the semantics were kind of awkward in CoffeeScript anyway, since the code basically just gets added in the closure where the class is constructed), so CoffeeScript classes can't very easily map to JS classes. However, I think it should be possible to make it work with some additional effort.

There are 16 files in my codebase that fail due to this error, and I've broken them down into two cases:
* In 7 of the files, class-level code was used to declare variables (usually constants, sometimes not) that were logically scoped to the class. In every one of these cases, it would be possible to pull the variable outside the class without introducing name collisions (although sometimes there was nontrivial code computing the variable, so it wouldn't be just a matter of moving a single line). In two of the cases, the "variables" were actually nested classes.
* In 9 of the files, the class-level code was used to assign properties to the constructor and sometimes the prototype.

I think it should be possible to implement a general-purpose transformation that always uses the `class` keyword. Here's an example that demonstrates many of the cases:

```coffee
class Foo
  SOME_CONSTANT = 7
  computeThing: -> SOME_CONSTANT + 3
  @prototype.abc = runExpensiveComputation()
  @def = 4
  computeOtherThing: -> @abc + Foo.def
```
would map to
```js
let SOME_CONSTANT;
class Foo {
  static initClass() {
    SOME_CONSTANT = 7;
    this.prototype.abc = runExpensiveComputation();
    this.def = 4;
  }
  computeThing() { return SOME_CONSTANT + 3; }
  computeOtherThing() { return this.abc + Foo.def; }
}
Foo.initClass();
```

Some notes on this approach:
* Rather than putting the code directly in the class body, it's put in a static method at the top of the class. This looks and behaves similarly to `static` init blocks in Java, but we need to explicitly call `initClass` after the end of the class. The meaning of `this` within a static method and within a CoffeeScript class is the same (it's the constructor), so this works out fairly well. Also, in most cases, I think it's most natural to put this code at the top.
* This involves reordering the class body so that the methods are all at the bottom and the non-methods are all at the top. In terms of execution order, the methods are all attached first, then the other code is run. This is generally safe because attaching a method doesn't rely on any state and doesn't have side effects. Technically this could break intermediate lines that assume that `this.prototype` does *not* yet have methods that are defined lower in the class, but that seems really obscure and hopefully wouldn't ever come up. If there is a reasonable case where code would do that, it might cause this approach to fall apart, but hopefully not.
* If the `initClass` name is taken, we'd need to pick a free name. FWIW, in the codebase I work in, that name is never used.
* Since `SOME_CONSTANT` needs to be used within the class, we need to make a variable for it somewhere, so we place a variable declaration outside the class. This is only safe if we can determine that `SOME_CONSTANT` is not already declared in the outer scope.
* If the `SOME_CONSTANT` name was already taken elsewhere, we could create a new scope by wrapping the class creation in an IIFE and putting `var SOME_CONSTANT` in the IIFE body. A bit ugly, but correct, and like I mentioned, this case never comes up in my codebase.
* It may be possible to do enough static analysis that `let SOME_CONSTANT = 7;` can be placed entirely outside the class. But that could get hard, and may not be worth the effort.
* If the class is in an expression context (which I believe doesn't come up in my code), we would need to wrap the class within an IIFE so that we can call `initClass` before returning it.

A related issue that I've run into (which I mentioned in #293) is that CoffeeScript class fields (which attach the value to the prototype) are converted into JavaScript class properties as defined in the [class fields](https://github.com/tc39/proposal-class-public-fields) stage 2 proposal, but the behavior is actually different enough that it introduces bugs for me. Backbone classes end up doing a lot of work in the constructor, including reading these class fields, but JavaScript class fields aren't set until after the constructor finishes. I've worked around this in my code by using the same `initClass` pattern above and setting values on the prototype, so if that pattern gets implemented for general class code, it could also be an alternate and more correct implementation of class fields (possibly behind a preference).

Feedback welcome, of course :-) . I have plenty of other things to do as well, but currently my thought is I'll get started on this at some point in the near future unless there are any objections or other ideas.

alangpierce:
Closes #397. See that issue for a much more thorough description of the strategy
here. This change doesn't handle all cases described in that issue (in
particular, it doesn't support arbitrary code within classes that are used in an
expression context), but it handles enough cases to support all classes in the
codebase I'm working with.

Since this transform significantly rearranges classes, I decided to do it as a
CS->CS transform in the normalize stage. That should hopefully avoid weird
magic-string issues in the main stage.

This approach is maybe a bit controversial in that it changes the implementation
for class fields to use prototype properties. In my case, putting class
properties on the prototype is necessary to avoid widespread correctness issues
in my codebase. (See http://benmccormick.org/2015/07/06/backbone-and-es6-classes-revisited/
for more details.) My current thought is I'll leave the class fields code in
place and we can make the behavior configurable with a command line flag if
there's demand for it. Or I can add the flag as part of this PR.

### DIFF
--- a/src/stages/normalize/index.js
+++ b/src/stages/normalize/index.js
@@ -1,4 +1,6 @@
 import BlockPatcher from './patchers/BlockPatcher.js';
+import ClassPatcher from './patchers/ClassPatcher.js';
+import AssignOpPatcher from './patchers/AssignOpPatcher.js';
 import ConditionalPatcher from './patchers/ConditionalPatcher.js';
 import ForInPatcher from './patchers/ForInPatcher.js';
 import ForOfPatcher from './patchers/ForOfPatcher.js';
@@ -47,6 +49,13 @@ export default class NormalizeStage extends TransformCoffeeScriptStage {
 
       case 'While':
         return WhilePatcher;
+
+      case 'Class':
+        return ClassPatcher;
+
+      case 'AssignOp':
+      case 'ClassProtoAssignOp':
+        return AssignOpPatcher;
 
       case 'Program':
         return ProgramPatcher;
--- a/src/stages/normalize/patchers/AssignOpPatcher.js
+++ b/src/stages/normalize/patchers/AssignOpPatcher.js
@@ -1,0 +1,12 @@
+import PassthroughPatcher from '../../../patchers/PassthroughPatcher.js';
+
+export default class AssignOpPatcher extends PassthroughPatcher {
+  key: NodePatcher;
+  expression: NodePatcher;
+
+  constructor(node: Node, context: ParseContext, editor: Editor, key: NodePatcher, expression: NodePatcher) {
+    super(node, context, editor, key, expression);
+    this.key = key;
+    this.expression = expression;
+  }
+}
--- a/src/stages/normalize/patchers/ClassPatcher.js
+++ b/src/stages/normalize/patchers/ClassPatcher.js
@@ -1,0 +1,218 @@
+import NodePatcher from './../../../patchers/NodePatcher.js';
+import AssignOpPatcher from './AssignOpPatcher.js';
+
+import type { Node, ParseContext, Editor } from './../../../patchers/types.js';
+
+export default class ClassPatcher extends NodePatcher {
+  nameAssignee: ?NodePatcher;
+  superclass: ?NodePatcher;
+  body: ?BlockPatcher;
+
+  constructor(node: Node, context: ParseContext, editor: Editor, nameAssignee: ?NodePatcher, parent: ?NodePatcher, body: ?BlockPatcher) {
+    super(node, context, editor);
+    this.nameAssignee = nameAssignee;
+    this.superclass = parent;
+    this.body = body;
+  }
+
+  /**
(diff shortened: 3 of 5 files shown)


alangpierce:
Ok, landing. As mentioned, I can add a preference to output JS class fields if there's demand for it.

PR #509, opened 2016-10-30
fix: improve the process of patching code for use elsewhere

### LINKED ISSUE #498: Crash when using `not of` in an IIFE-style loop

```coffee
a = (b for b of c when d not of e)
```

[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=a%20%3D%20(b%20for%20b%20of%20c%20when%20d%20not%20of%20e))

I get this error:

```
AddVariableDeclarationsStage failed to parse: Unexpected token (4:15)
  2 |   result = [];
  3 |   for (b in c) {
> 4 |     if (d in e)) {;
    |                ^
  5 |       result.push(b);
  6 |     };
```

This is similar to #491, but has a different underlying cause.

alangpierce:
Fixes #498

Previously, you might call `patch` on a sub-patcher, then use `slice` to get the
generated code, but this didn't always work out well, since we always do
`insertLeft` (for magic-string sanity), so slice would sometimes miss an
open-paren inserted at the start.

This commit changes the loop filter case to use this new method, but probably
other code paths should be moved over to it as well in the future.

### DIFF
--- a/src/patchers/NodePatcher.js
+++ b/src/patchers/NodePatcher.js
@@ -233,6 +233,28 @@ export default class NodePatcher {
   }
 
   /**
+   * Patch the given expression and get the underlying generated code. This is
+   * more robust then calling patch and slice directly, since it also includes
+   * code inserted at contentStart (which normally isn't picked up by slice
+   * because it's inserted to the left of the index boundary). To accomplish
+   * this, we look at the range from contentStart - 1 to contentStart before and
+   * after patching and include anything new that was added.
+   */
+  patchAndGetCode(options={}) {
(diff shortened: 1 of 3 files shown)


src/patchers/NodePatcher.js
```diff
@@ -233,6 +233,28 @@ export default class NodePatcher {
   }
 
   /**
+   * Patch the given expression and get the underlying generated code. This is
+   * more robust then calling patch and slice directly, since it also includes
```
eventualbuddha: Silly thing to point out, but this should be "than" instead of "then".

alangpierce: Oops :-)

PR #519, opened 2016-11-02
ESlint fix

Eden-Sun:


package.json
```diff
@@ -39,6 +39,10 @@
     "decaffeinate-coffeescript": "^1.10.0-patch5",
     "decaffeinate-parser": "^2.0.2",
     "detect-indent": "^4.0.0",
+    "eslint-config-standard": "^6.2.1",
+    "eslint-plugin-html": "^1.5.5",
+    "eslint-plugin-promise": "^3.3.0",
+    "eslint-plugin-standard": "^2.0.1",
```
eventualbuddha: Why are all these needed?

Eden-Sun: "extends": "eslint:recommended"

it means to use esint-config-standard
and it depends those 3 plugins

Eden-Sun: Didn't  you guys run `npm run lint` without errors?

.eslintrc
```diff
@@ -6,12 +6,12 @@
     "mocha": true
   },
   "rules": {
-    "semi": 2,
+    "semi": [2, "always"],
```
alangpierce: According to http://eslint.org/docs/rules/semi#options , "always" is the default, so I think this change doesn't do anything.

alangpierce:
I don't get any errors when I run `npm run lint`. @Eden-Sun, if you're seeing failures, maybe you have an .eslintrc file in some directory above the decaffeinate directory? I think that might cause eslint to apply additional rules that aren't in the decaffeinate lint config. If that's the problem, then I think a way to avoid it is to add `"root": true` to the decaffeinate eslintrc ([docs here](http://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy)).

Eden-Sun:
Yes, it was my mistake due to root eslint config.

Issue #322, opened 2016-07-17
Wrong behavior when using a range with "by" in a "for in" loop


alangpierce:
It looks like when you loop over a range and use the `by` keyword, that forces the direction of the iteration, rather than the direction being decided based on whether the range goes forward or backward. This is useful to avoid a gotcha where iterating over `[1..0]` ends up iterating over 1 and then 0. However, decaffeinate ignores this behavior and just computes the range first, then loops over it like normal.

Here's a simple example:

``` coffee
for a in [1..0] by 1
  b()
```

([decaffeinate repl](http://decaffeinate-project.org/repl/#?evaluate=true&code=for%20a%20in%20%5B1..0%5D%20by%201%0A%20%20b%28%29))
([coffeescript repl](http://coffeescript.org/#try:for%20a%20in%20%5B1..0%5D%20by%201%0A%20%20b%28%29))

The correct behavior is to do nothing, but the decaffeinate output loops over 1 and then 0.

Here's a more realistic example with the same problem (which is in the same form as some code I actually ran into):

``` coffee
for a in [1..len-1] by 1
  b()
```

([decaffeinate repl](http://decaffeinate-project.org/repl/#?evaluate=true&code=for%20a%20in%20%5B1..len-1%5D%20by%201%0A%20%20b%28%29))
([coffeescript repl](http://coffeescript.org/#try:for%20a%20in%20%5B1..len-1%5D%20by%201%0A%20%20b%28%29))

eventualbuddha:
Related, decaffeinate shouldn't even generate the range in most of these cases.

eventualbuddha:
@alangpierce can you confirm this is working for you now?

Eden-Sun:
It does.

alangpierce:
Yep, looks good to me! Closing.

PR #547, opened 2016-11-16
feat: use `for(;;)` style loops for some `for [a..b]` style loops

eventualbuddha:
This commit does not use `for(;;)` when the loop has a dynamic range unless it also has an explicit step. That means `for [a..b]` will become a `for…of` loop. I added two skipped tests for turning them into `for(;;)` style, but it would require a fairly verbose loop header. However, that may still be better than requiring the `__range__` helper.

### DIFF
--- a/test/for_test.js
+++ b/test/for_test.js
@@ -1,7 +1,7 @@
 import check from './support/check.js';
 import validate from './support/validate.js';
 
-describe('for loops', () => {
+describe.only('for loops', () => {
   it('transforms basic for-of loops into for-in', () => {
     check(`
       for k of o
@@ -174,7 +174,7 @@ describe('for loops', () => {
       for [0..1]
         2
     `, `
-      for (let value of [0, 1]) {
+      for (let i = 0; i <= 1; i++) {
         2;
       }
     `);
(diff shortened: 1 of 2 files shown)


eventualbuddha:
I believe this will fix #322.

test/for_test.js
```diff
@@ -1,7 +1,7 @@
 import check from './support/check.js';
 import validate from './support/validate.js';
 
-describe('for loops', () => {
+describe.only('for loops', () => {
```
alangpierce: Looks like you left the `only` in by mistake.

alangpierce:
Looks good to me, although you accidentally disabled all of the other tests. :-)

Kind of sad that we can't do this transform for dynamic ranges. I have a lot of code like `for i of [0...len]`, where a human could obviously see that it means to count up. Maybe there could be a loose mode that has a heuristic like "if the range starts at the constant 0, patch as if the loop is counting up". Fortunately, this change makes it pretty easy to add those types of optional heuristics if we want. Or I could just go through code beforehand and write `by 1` when I'm confident that it's counting up.

I feel like I would pretty much never write code that decides at runtime whether to count up or down, so I maybe even prefer `__range__` over CoffeeScript's output because it makes it more obvious that someone should go in and write it the right way.

test/for_test.js
eventualbuddha: Eek! Fixed.

Issue #557, opened 2016-11-21
Rework string handling across the decaffeinate projects


alangpierce:
This is more of a meta-issue for a refactor I'm planning to do.

String (and regex) handling in CoffeeScript has a number of different cases, and while decaffeinate handles many of them, there are some incomplete and incorrect cases. It looks like decaffeinate's implementation is sort of awkwardly split across decaffeinate, decaffeinate-parser, and coffee-lex, and would be good to rework those details. I'm using this issue to plan and organize that work.

This issue should address all existing string/interpolation bugs (or at least make them much easier to address): #556, #554, #406, and #341.

As far as I understand, CoffeeScript has three algorithms for removing whitespace in multiline strings:

* **Multiline strings:** In multiline single-quoted and double-quoted strings, the only text to consider from each line is the range from the first non-whitespace character to the last non-whitespace character. These lines are then split by spaces.
* **Herestrings:** When a herestring spans multiple lines, the common leading indentation from each line is removed, plus lots of other subtle details in how the first and last line are treated. I documented many of these details in the documentation for [parseHerestring](https://github.com/decaffeinate/decaffeinate-parser/blob/master/src/util/parseLiteral.js#L53).
* **Heregexes:** Whether multiline or not, ALL whitespace is removed unless it is escaped. Also, `#` comments are respected as long as they are immediately preceded by a space, tab, or newline, and they last until the end of the line.

In all three cases, string interpolations are allowed (except when using single-quotes in normal strings or herestrings), and the rules are the same.

Here's how these cases are currently handled (or not) in decaffeinate:

* For multiline strings, coffee-lex has a special case for them in `stringLocationsFromStream` to handle interpolations, although it doesn't actually determine which whitespace to remove. When there are no interpolations, coffee-lex just emits a single `DSTRING` or `SSTRING` token. decaffeinate-parser emits either a `String` or a `TemplateLiteral` node based on whether there are any interpolations. It does a lot of string parsing work in the `parseLiteral` function, but doesn't compute whitespace to remove. On the decaffeinate side, `String` nodes have the whitespace removal calculated in `parseMultilineString`. `TemplateLiteral` nodes don't properly remove whitespace, causing #554.
[...]

PR #572, opened 2016-11-27
fix: update decaffeinate to use new string lexing/parsing

### LINKED ISSUE #406: Escaping spaces before new-line in multi-line string not supported

``` coffee
a = "a\
b"
```

http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=a%20%3D%20%22a%5C%0Ab%22

I get this output:

``` js
let a = "a\ b";
```

``` js
let a = "a" 
+ "b"
```

### LINKED ISSUE #554: Multiline double-quoted strings with interpolations are converted incorrectly
(Originally filed in #406, moved here.)

```coffee
"
a
#{b}
c
"
```

[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=%22%0Aa%0A%23%7Bb%7D%0Ac%0A%22)

I get this output:

```js
`
a
${b}
c
`;
```

Either
```js
`a ${b} c`
```

or
```js
`\
a \
${b} \
c\
`
```

CoffeeScript gives this:
```js
"a " + b + " c";
```

The issue here is that multiline double-quoted strings and herestrings have different rules for how to remove whitespace, but when there's an interpolation, decaffeinate-parser just uses the `TemplateLiteral` node type, and `TemplateLiteralPatcher` in decaffeinate doesn't handle this case (and it would be hard for it to do so with the AST information it has).

### LINKED ISSUE #556: Herestrings with interpolations don't properly handle leading whitespace

```coffee
"""    a
b#{c}
"""
```

[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=%22%22%22%20%20%20%20a%0Ab%23%7Bc%7D%0A%22%22%22)

I get this output:

```js
`a
b${c}`;
```

```js
`    a
b${c}`;
```

CoffeeScript gives this output:

```js
"    a\nb" + c;
```

Note that removing the string interpolation at the end fixes the issue. This behavior is really a quirky edge case in the CoffeeScript lexer, but the more fundamental issue is that decaffeinate uses two different code paths for removing indentation in herestrings with interpolations and herestrings without interpolations. Without interpolations, it uses `parseHerestring` in `parseLiteral.js` in decaffeinate-parser. With interpolations, it uses `stripSharedIndent` in decaffeinate. In https://github.com/decaffeinate/decaffeinate-parser/pull/44, I documented all of the subtle details that I could find, and `stripSharedIndent` follows some but not all of those details, and this is one case where they differ.

The right fix for this, I think, is to rework decaffeinate's string handling so that this type of logic is in one place regardless of whether or not there are any interpolations.

alangpierce:
Progress toward #557
Fixes #556
Fixes #554
Fixes #406

This change replaces TemplateLiteralPatcher, HerestringPatcher, and
StringPatcher with a new StringPatcher that handles all cases and uses the token
stream from coffee-lex.

One change from the previous behavior is that rather than removing newlines in
multiline strings, I now escape them. That way they stay multiline but have some
extra backslashes in the source code, which probably matches the intended
formatting better (even though it looks ugly in some cases).

Also fix a case in ObjectInitialiserMemberPatcher that was assuming that you
only need square brackets around object keys when the expression is not
repeatable. Instead, specifically check if the key will become a template
string, and handle the repeatable and non-repeatable cases separately.

In addition to making sure the tests pass, I also ran this change on the codebase
I'm trying to convert to JS and it didn't introduce any new errors.


src/stages/main/patchers/StringPatcher.js
```diff
@@ -1,67 +1,120 @@
...
-        if (line.prev.empty) {
-          patcher.remove(line.start, line.end);
-        } else {
-          patcher.overwrite(line.start, line.end, ' ');
+  /**
+   * Handle "padding" characters: characters like leading whitespace that should
+   * be removed according to the lexing rules. In addition to STRING_PADDING
+   * tokens, which indicate that the range should be removed, there are also
+   * STRING_LINE_SEPARATOR tokens that indicate that the newlines should be
+   * replace with a space.
```
eventualbuddha: replace ➡️ replaced

PR #575, opened 2016-11-28
feat: add support for heregex templates

### LINKED ISSUE #341: Support string interpolation within heregexes
Example:

``` coffee
a = ///#{b}///
```

[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&code=a%20%3D%20%2F%2F%2F%23%7Bb%7D%2F%2F%2F)

Currently it crashes with the error "Cannot read property '2' of null".

### LINKED ISSUE #557: Rework string handling across the decaffeinate projects

alangpierce:
Fixes #557
Fixes #341

Most of the existing code in StringPatcher was also relevant for heregexes, so I
pulled the shared code into an InterpolatedPatcher class.

One oddity here is the way backslashes are handled in heregexes. You can use a
backslash to escape a whitespace character (i.e. to include it), but other than
that, backslashes just represent the backslash character. So when turning a
heregex into a string, we need to escape backslashes when not followed by
whitespace, but skip them if they are. To accomplish this, I generalized the
`escape` function to parameterize how it skips characters.

With this change, decaffeinate can convert all 1213 files in my codebase!


eventualbuddha:
> With this change, decaffeinate can convert all 1213 files in my codebase!

🎉🎉

But do the tests pass? 😉

alangpierce:
Heh, nope, still definitely some test failures. Some seem hard to fix in decaffeinate, like using `this` before `super` in a constructor and some differences between `import` and `require`. But still, there's definitely progress to be made and maybe creative solutions for problems that seem hard now.

PR #593, opened 2016-12-04
feat: add an option to prefer const rather than let

### LINKED ISSUE #408: Make using `const` when possible an option
We should add a `--prefer-const` flag to the command line which translates to a `preferConst` option.

alangpierce:
Fixes #408.

While I'm at it, I made the CLI exit with an error message if it sees an invalid
argument. We've dropped support for node 0.12, so it's safe to use `startsWith`,
but the conversion guide hadn't been updated, so I also updated that.

### DIFF
--- a/docs/conversion-guide.md
+++ b/docs/conversion-guide.md
@@ -7,7 +7,7 @@ JavaScript using decaffeinate.
 
 This guide assumes a few things:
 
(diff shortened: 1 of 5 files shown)


eventualbuddha:
I've been meaning to do this for a while. Thanks!

PR #608, opened 2016-12-11
fix: properly detect break statements within switch cases

### LINKED ISSUE #607: Switch case with while loop does not have break statement properly inserted

```coffee
switch a
  when 'b'
    while false
      break
  when 'c'
    console.log 'Got here'
```

[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=switch%20a%0A%20%20when%20'b'%0A%20%20%20%20while%20false%0A%20%20%20%20%20%20break%0A%20%20when%20'c'%0A%20%20%20%20console.log%20'Got%20here')

I get this output:

```js
switch (a) {
  case 'b':
    while (false) {
      break;
    }
  case 'c':
    console.log('Got here');
    break;
}
```

```js
switch (a) {
  case 'b':
    while (false) {
      break;
    }
    break;
  case 'c':
    console.log('Got here');
    break;
}
```

alangpierce:
Fixes #607.

Previously, we would detect if the switch case body ends with a `break` token,
but that's actually not good enough; we should directly check that the last
statement is a break statement.

### DIFF
--- a/src/stages/main/patchers/SwitchCasePatcher.js
+++ b/src/stages/main/patchers/SwitchCasePatcher.js
@@ -1,6 +1,7 @@
+import BlockPatcher from './BlockPatcher.js';
 import NodePatcher from '../../../patchers/NodePatcher.js';
 import type { PatcherContext, SourceToken } from '../../../patchers/types.js';
-import { BREAK, COMMA, THEN, WHEN } from 'coffee-lex';
+import { COMMA, THEN, WHEN } from 'coffee-lex';
 
 export default class SwitchCasePatcher extends NodePatcher {
   conditions: Array<NodePatcher>;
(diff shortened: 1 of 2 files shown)


src/stages/main/patchers/SwitchCasePatcher.js
```diff
@@ -139,13 +139,14 @@ export default class SwitchCasePatcher extends NodePatcher {
...
-    } else {
-      return null;
+  hasExistingBreak(): boolean {
+    if (!(this.consequent instanceof BlockPatcher)) {
+      return false;
     }
+    let lastStatement = this.consequent.statements[this.consequent.statements.length - 1];
+    return lastStatement &&
+        lastStatement.node.type === 'Identifier' &&
+        lastStatement.node.data === 'break';
```
eventualbuddha: Related, it seems wrong that this is an identifier and not a `BreakStatement` or something.

Issue #107, opened 2016-01-06
Support non-terminal rest params


eventualbuddha:
CoffeeScript allows this:

``` coffee
(keys..., callback) ->
```

But you can't simply turn that into this:

``` js
(function(...keys, callback) {});
```

Because rest params have to be at the end. We'll have to do some juggling to support this.

alangpierce:
[Example repl link](http://decaffeinate-project.org/repl/#?evaluate=false&stage=full&code=(args...%2C%20funcArg)%20-%3E%20new%20Map%20%7B%7D)

PR #646, opened 2016-12-28
feat: implement non-terminal rest in array destructures

alangpierce:
Progress toward #107

We already had code to handle expansion nodes, so non-terminal rest is almost
the same, we just need to produce one more assignment that slices the middle
part of the array.

I also reworked some of the code to use `patchAndGetCode` when possible and to
patch code in order as much as we can.

The plan is to implement non-terminal rest and expansion for parameters by moving
the trailing parameters to an array destructure in the normalize step, so we'll
only need one code path that needs to think about the details of generating this
type of code.

### DIFF
--- a/src/stages/main/patchers/AssignOpPatcher.js
+++ b/src/stages/main/patchers/AssignOpPatcher.js
@@ -1,6 +1,7 @@
 import ArrayInitialiserPatcher from './ArrayInitialiserPatcher';
 import ExpansionPatcher from './ExpansionPatcher';
 import NodePatcher from './../../../patchers/NodePatcher';
+import SpreadPatcher from './SpreadPatcher';
 import type { PatcherContext } from './../../../patchers/types';
 
 export default class AssignOpPatcher extends NodePatcher {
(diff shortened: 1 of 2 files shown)


src/stages/main/patchers/AssignOpPatcher.js
```diff
@@ -133,15 +131,34 @@ export default class AssignOpPatcher extends NodePatcher {
...
-      if (i < expansionIndex) {
-        key = `${i}`;
+        if (assignee instanceof ExpansionPatcher) {
+          // Don't patch this node, since we'll just remove it. We know there's
+          // an assignee after the expansion because otherwise we would have
+          // returned above.
+          this.remove(assignee.outerStart, assignees[i + 1].outerStart);
+        } else if (assignee instanceof SpreadPatcher) {
+          // Don't patch the spread itself since the new leading "..." can be a
+          // hassle and wont' be used anyway. Instead, just patch the underlying
```
eventualbuddha: typo: "won't"

src/stages/main/patchers/AssignOpPatcher.js
```diff
@@ -133,15 +131,34 @@ export default class AssignOpPatcher extends NodePatcher {
...
+          // expression and get its code.
+          let assigneeCode = assignee.expression.patchAndGetCode();
+          let sliceEnd = `${arrReference}.length - ${assignees.length - i - 1}`;
+          this.overwrite(
+            assignee.outerStart, assignee.outerEnd,
+            `${assigneeCode} = ${arrReference}.slice(${i}, ${sliceEnd})`
+          );
+        } else {
+          throw new Error('Unexpected expansion node type.');
```
eventualbuddha: `throw this.error(…)`, right?

alangpierce: Oops, yeah you're right.

PR #677, opened 2016-12-30
fix: ensure inline loops work as expected

### LINKED ISSUE #674: Loop statements with an inline body no longer work

```coffee
loop a
```

[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=loop%20a)

I get this error:

```
MainStage failed to parse: unexpected identifier
> 1 | while true a
    |            ^
```

@eventualbuddha looks like this was caused by https://github.com/decaffeinate/decaffeinate/commit/4178ee0c4edc3a8d35bdfb1e5ced91afdbd199

I think in the inline form, it needs to convert to `while true then a`. Another case that needs to work correctly is `loop then a`, so I think the most straightforward fix is to check if the `then` is missing and add it if so.

eventualbuddha:
Fixes #674

### DIFF
--- a/test/while_test.js
+++ b/test/while_test.js
@@ -78,6 +78,14 @@ describe('while', () => {
     `);
   });
 
+  it('turns `loop` with an inline body into an inline `while` loop with an always-true condition', () => {
+    check(`
+      loop a
+    `, `
+      while (true) { a; }
+    `);
+  });
+
   it('handles `while` loops used as an expression', () => {
     check(`
       a(b while c)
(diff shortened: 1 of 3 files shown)


test/while_test.js
```diff
@@ -78,6 +78,14 @@ describe('while', () => {
     `);
   });
 
+  it('turns `loop` with an inline body into an inline `while` loop with an always-true condition', () => {
+    check(`
+      loop a
```
alangpierce: Can you also add a test for `loop then a`?

eventualbuddha: We actually [already have one](https://github.com/decaffeinate/decaffeinate/blob/1e38f68f727674065fe15a9e8b88f8fa518b3cb7/test/while_test.js#L208-L214).

alangpierce: Oh, nevermind, sounds good then.

PR #693, opened 2016-12-31
cleanup: remove the old (unused) regex patcher

eventualbuddha:


eventualbuddha:
I went with `Regex` for the node, so `RegexPatcher` seems good to me.

PR #697, opened 2017-01-01
feat: implement compound modulo assignment

### LINKED ISSUE #696: Implement compound modulo assignment

```coffee
a %%= b
```

[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=a%20%25%25%3D%20b)

I get this error:

```
AddVariableDeclarationsStage failed to parse: Unexpected token (1:3)
> 1 | a %%= b;
    |    ^
```

From https://github.com/jashkenas/coffeescript/blob/master/test/operators.coffee

alangpierce:
Fixes #696

We already parsed this correctly, we just needed a new patcher type for it. I
also pulled out the helper itself into a shared util function.

### DIFF
--- a/test/compound_assignment_test.js
+++ b/test/compound_assignment_test.js
@@ -29,6 +29,33 @@ describe('compound assignment', () => {
     check(`a ^= 1`, `a ^= 1;`);
   });
 
+  it('handles compound modulo assignment', () => {
+    check(`
+      a %%= b
+    `, `
+      let a = __mod__(a, b);
+      function __mod__(a, b) {
+        a = +a;
+        b = +b;
+        return (a % b + b) % b;
+      }
+    `);
+  });
+
+  it('handles compound modulo assignment with a non-repeatable LHS', () => {
+    check(`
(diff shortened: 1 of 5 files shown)


test/compound_assignment_test.js
```diff
@@ -29,6 +29,33 @@ describe('compound assignment', () => {
     check(`a ^= 1`, `a ^= 1;`);
   });
 
+  it('handles compound modulo assignment', () => {
+    check(`
+      a %%= b
+    `, `
+      let a = __mod__(a, b);
```
eventualbuddha: This seems like an issue with add-variable-declarations, and possibly also esnext. Assignment to a binding that is also referenced in the rvalue should probably assume a global. Also, even if there is such a `var`, turning it into a `let` introduces a TDZ issue.

alangpierce: Good point, I didn't think of that. Hopefully any reasonable code wouldn't rely on that behavior, but yeah, would be good to address eventually.

PR #707, opened 2017-01-02
feat: implement splice operations

### LINKED ISSUE #668: Add support for splice operations

```coffee
a[b..c] = d
```

[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=a%5Bb..c%5D%20%3D%20d)

I get this error:

```
Invalid left-hand side in assignment expression (1:0)
```

The CoffeeScript compiler gives this code:

```js
[].splice.apply(a, [b, c - b + 1].concat(d)), d;
```

alangpierce:
Fixes #668

This is heavily based off of the slice patching code, with some modifications.
It's also a little strange in that the slice patcher patches only part of the
slice call and the assign op patcher patches the rest, but that seemed like the
cleanest way to do it with the way things are structured now.

### DIFF
--- a/test/support/validate.js
+++ b/test/support/validate.js
@@ -2,7 +2,7 @@ import * as babel from 'babel-core';
 import * as vm from 'vm';
 import { compile } from 'decaffeinate-coffeescript';
 import { convert, PatchError } from '../..';
-import { strictEqual } from 'assert';
+import { deepEqual } from 'assert';
 
 /**
  * validate takes coffee-script as input with code that sets the variable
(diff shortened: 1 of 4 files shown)


alangpierce:
Oh, actually, looks like you can use splice to insert individual elements, e.g. `a[2...4] = 'foo'` behaves the same as `a[2...4] = ['foo']`, since they both use `Array.prototype.concat`. I'll need to think a little more about how to do that correctly. Seems like maybe the best approach is to just mimic CoffeeScript and use `concat`.

alangpierce:
Oh, nevermind, looks like `Array.from` has the same behavior, and I think I prefer it to `...[].concat(`. I'll push two more tests that use `verify` and this should be good to go after that.

test/support/validate.js
```diff
@@ -48,7 +48,7 @@ function runValidation(source: string, expectedOutput: ?any) {
   let coffeeOutput = runCodeAndExtract(coffeeES5);
   let decaffeinateOutput = runCodeAndExtract(decaffeinateES5);
   try {
-    strictEqual(decaffeinateOutput, coffeeOutput);
```
eventualbuddha: Why is this change needed?

alangpierce: Previously, all usages of `validate` expected the value of `o` at the end to be a primitive, so an `===` check was good enough in the assertion. I added the first case where the expected result is an array, so we need to do a deeper check of the contents.

PR #725, opened 2017-01-06
build: add an improved update-website script

eventualbuddha:


script/update-website.ts
```diff
@@ -0,0 +1,137 @@
+#!/usr/bin/env ts-node
+
+import getLatestVersion from './getLatestVersion';
+import { execFile } from 'child_process';
```
alangpierce: Did you consider https://github.com/normalize/mz ? That's what I use in bulk-decaffeinate and it means I almost never need to wrap things in `new Promise`

eventualbuddha: I think I'd seen that before, but never dug into it. Thanks for the pointer!

PR #735, opened 2017-01-07
refactor: convert a few utilities and tests to TypeScript

eventualbuddha:


src/utils/determineIndent.ts
```diff
@@ -1,11 +1,9 @@
-/* @flow */
-
-import detectIndent from 'detect-indent';
+import * as detectIndent from 'detect-indent';
```
alangpierce: I think you don't want `import *` here?

PR #737, opened 2017-01-07
feat: more intelligently handle invalid constructors

### LINKED ISSUE #591: Resulting JS can incorrectly have this before super in a constructor

```coffee
class A extends B
  constructor: ->
    @c = d
    super
```

[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=class%20A%20extends%20B%0A%20%20constructor%3A%20-%3E%0A%20%20%20%20%40c%20%3D%20d%0A%20%20%20%20super)

I get this output:

```js
class A extends B {
  constructor() {
    this.c = d;
    super(...arguments);
  }
}
```

Here's a [babel repl link](http://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&code=class%20A%20extends%20B%20%7B%0A%20%20constructor()%20%7B%0A%20%20%20%20this.c%20%3D%20d%3B%0A%20%20%20%20super(...arguments)%3B%0A%20%20%7D%0A%7D) that shows the error here: "'this' is not allowed before super()".

I don't have any great ideas on how decaffeinate can handle this, but it affects 11 files in my codebase, so it seems useful to address or at least acknowledge the problem. Here are some possible ideas:
* Maybe decaffeinate could detect this case and generate CS-style classes instead of JS classes, so that it's allowed to use `this` before `super`. One downside here, though, is that JS and CS classes have interop issues, so it's best to use only JS classes if possible.
* There may be some clever way to rewrite the code to be correct. My best idea at the moment in the above code example is to change `c` into a getter, which I think means that it will be accessible to the super constructor, but obviously the general case is likely much harder.
* We could document that this is an issue and provide suggestions on how to fix it.
* decaffeinate could detect this case and give an error or a warning so that people are aware of it earlier rather than running into a babel failure after decaffeinate is finished.

### LINKED ISSUE #726: Default value in constructor breaks compilation

```coffee
class Nukes
  constructor: (y = 'foo') ->
  launch: (x) => true
```

[repl](http://decaffeinate-project.org/repl/#?evaluate=true&stage=EsnextStage&code=class%20Nukes%0A%20%20constructor%3A%20(y%20%3D%20'foo')%20-%3E%0A%20%20launch%3A%20(x)%20%3D%3E%20true)

I get this error:

```
AddVariableDeclarationsStage failed to parse: Unexpected token, expected ) (3:44)
  1 | class Nukes {
  2 |   constructor(y) {
> 3 |     if (this.launch = this.launch.bind(this);
    |                                             ^
  4 |     y == null) { y = 'foo'; };
  5 |   }
```

alangpierce:
Fixes #591
Fixes #726

A major difference between CoffeeScript and JavaScript class constructors is
that JavaScript class constructors require you to call `super` and require it
to be called before the first use of `this` (when the class is a subclass).
This has caused two significant issues in decaffeinate:
* CS code that uses `this` before `super` or omits `super` will convert to code
  that is rejected by babel. This happened for 11 files in my codebase.
* Subclasses that use bound methods will put the binding code after the `super`
  call, while the CoffeeScript semantics are to put it before the `super` call.
  This caused (at least) two correctness issues in my codebase, only one of
  which was caught the tests. (Both of these were backbone subclasses where the
  `initialize` method gets called before the super call ends.)

Since there doesn't seem to be a great answer to this problem, this commit
provides three options for how to handle it:
* Give an error rather than generating incorrect code. This is the default
  behavior. If a constructor uses `this` before `super` or omits `super`, or if
  a subclass uses a bound method, then decaffeinate gives an error for the file
  explaining the issue and the different options. This flags the issue for
  people not aware of it, and gives people a chance to change their CoffeeScript
  code to not have these issues (e.g. rewriting constructors and/or changing
  bound methods to explicit bindings).
* `--allow-invalid-constructors` generates JS code even if `this` is used before
  `super` in the constructor, and it generates code that would be correct if JS
  allowed it (method bindings at the start of the constructor). Technically, this
  code is syntactically valid, but will crash at runtime in a JS interpreter and
  will be rejected by babel. The intention here is that it may be useful to
  convert the code anyway and fix the invalid constructors later, e.g. moving
  the binding lines after doing a sanity check.
* `--enable-babel-constructor-workaround` generates the same invalid code, but
  also puts a code snippet at the top of each constructor that tricks babel into
  allowing `this` before `super` in constructors (even at runtime). This works
  because babel doesn't use fully-compliant JS classes, and it's certainly a
  hack that might break due to a change in babel, but as far as I can tell, it's
  completely correct as long as you're using babel. This is especially useful as
  a way to automatically convert a project and run all of the tests, but
  probably shouldn't be used in production or at least should be removed as soon
  as possible if it is.

Notably, there is no longer a way to put method bindings after the super call.
This both makes the implementation simpler and is more in line with "either
generate correct code or give an error". There could be an option for this in
the future, but really, I think it's best for people to manually check that
reordering the constructor is ok.

Another possible future option is to generate classes the same way that
CoffeeScript does, or do some other less hacky trick to allow `this` before
`super`. But JS classes fundamentally don't allow this, and CS classes can't
extend JS classes, so it seems like the only 100% safe way to do this is to
NEVER generate JS classes, which seems like it would be more work and defeat a
lot of the purpose of decaffeinate.

### DIFF
--- a/src/stages/main/patchers/ConstructorPatcher.js
+++ b/src/stages/main/patchers/ConstructorPatcher.js
@@ -1,5 +1,9 @@
+import ClassPatcher from './ClassPatcher';
 import ObjectBodyMemberPatcher from './ObjectBodyMemberPatcher';
+import babelConstructorWorkaroundLines from '../../../utils/babelConstructorWorkaroundLines';
+import getInvalidConstructorErrorMessage from '../../../utils/getInvalidConstructorErrorMessage';
 import traverse from '../../../utils/traverse';
+import { isFunction } from '../../../utils/types';
 import type FunctionPatcher from './FunctionPatcher';
 import type NodePatcher from '../../../patchers/NodePatcher';
 import type { PatcherContext } from './../../../patchers/types';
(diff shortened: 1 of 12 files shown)


alangpierce:
@eventualbuddha I imagine this may be a bit controversial, but let me know what you think. I've found `--enable-babel-constructor-workaround` to be a nice way to check that tests pass after decaffeinate, even if the generated code is really hacky. It's also nice in that it's a self-contained snippet that doesn't add much complexity to decaffeinate.

src/stages/main/patchers/ConstructorPatcher.js
```diff
@@ -73,6 +144,32 @@ export default class ConstructorPatcher extends ObjectBodyMemberPatcher {
...
+      let usesThis = false;
+      traverse(statements[i].node, child => {
+        if (usesThis) {
+          // Already found it, skip this one.
+          return false;
+        } else if (child.type === 'This') {
+          // Found it.
+          usesThis = true;
+        } else if (child.type === 'Class') {
+          // Don't go into other classes.
```
eventualbuddha: Or non-arrow functions?

alangpierce: Good point. Actually, I'll skip all functions to be consistent with babel's static analysis (we don't know when the function will be called).

balakumarshanmugam:
@alangpierce We are facing the same issue. In CS class, there is no constructor. But the CS class has bound functions. Can I convert that to JS by using "enable-babel-constructor-workaround"? It creates a constructor in JS class that has a block of code to trick babel into allowing this before super, then the method binding, then the super call. Is it fine to convert my CS file to JS?

alangpierce:
@balakumarshanmugam I would recommend changing your code so you don't need the workaround. For example, you can change this:

```coffee
class A extends B
  c: => 'Hello world'
```

to this:
```coffee
class A extends B
  constructor: ->
    super
    @c = @c.bind(this)
  
  c: -> 'Hello world'
```

But keep in mind that this changes the behavior of the code. In the second code snippet, the method binding only happens after the `super` call finishes. Usually this isn't a big deal, but there were a few cases in my codebase where moving it after `super` caused problems.

The `--enable-babel-constructor-workaround` option should be correct as long as you're using the current version of babel (it might break for future babel versions because it relies on the implementation details of babel, and it will definitely break if you move off of babel). So you're welcome to use that if you want, and it's especially useful for verifying that your tests pass after decaffeinate, but it's probably not code you want in your codebase long-term.

balakumarshanmugam:
@alangpierce Thanks for your response.

I tried the option which you gave as example. i.e. calling super before binding the functions. But like you did, I also faced problem because the class B is a backbone view and it expects the functions to be bound. This means that I have to bind the functions before calling super. So, for now, I used the --enable-babel-constructor-workaround option and it passes the test after decaffeinate.

alangpierce:
@balakumarshanmugam that makes sense. One thing that usually works for me is to bind the function at the start of `initialize`, which is called during the `super` call. Of course, you're welcome to do what you want with your code, but I think it's best to avoid `--enable-babel-constructor-workaround` as a long-term solution. But certainly one thing you could do is run with `--enable-babel-constructor-workaround` so you're in JavaScript, then figure out how to get rid of the babel hack when you get a chance at some point in the future.

PR #764, opened 2017-01-15
fix: better build soaked member access

### LINKED ISSUE #759: existential fail on for valAssignee

```coffee
i + j for j, i in foo?.bar ? [] 
```

[repl broken](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=i%20%2B%20j%20for%20j%2C%20i%20in%20foo%3F.bar%20%3F%20%5B%5D%20%0D%0A)

I get this error:

```
AddVariableDeclarationsStage failed to parse: Unexpected token, expected ; (1:26)
> 1 | iterable = foo, x => x.bar) != null ? __guard__(foo, x => x.bar) : []
    |                           ^
  2 | for (i = 0; i < iterable.length; i++) { j = iterable[i]; i + j;; }
  3 | 
```

simple explicit precedenace solves this:
```coffee
i + j for j, i in (foo?.bar) ? [] 
```
[repl working](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=i%20%2B%20j%20for%20j%2C%20i%20in%20(foo%3F.bar)%20%3F%20%5B%5D%20%0D%0A)

refack:
alt (#761)
Fixes #759

* might consider promoting new method to NodePatcher, might be better
  when prepending tokens.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules
 dist
 npm-debug.log
+.idea/
--- a/src/stages/main/patchers/SoakedMemberAccessOpPatcher.js
+++ b/src/stages/main/patchers/SoakedMemberAccessOpPatcher.js
@@ -29,13 +29,37 @@ export default class SoakedMemberAccessOpPatcher extends MemberAccessOpPatcher {
       let memberNameToken = this.getMemberNameSourceToken();
       this.overwrite(this.expression.outerEnd, memberNameToken.start, `, ${varName} => ${prefix}${varName}.`);
 
-      soakContainer.insert(soakContainer.contentStart, '__guard__(');
+      if (soakContainer.prepend)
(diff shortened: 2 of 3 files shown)


refack:
I understand 👍 
My main point was "correct" string manipulation, but if you'd rather handle it from the parent with `patchAndGetCode` it's totally makes sense.

refack:
Closing this PR (with no merge)

alangpierce:
Yeah, it's possible that the `prepend` approach could be made to work (although the name is actually a little weird because it puts the code on the right side of the index boundary), but at the very least it would require changing all inserts in the decaffeinate codebase to use `prepend` when it makes sense, or else you could get out-of-order inserts, and I suspect that various other problems would come up when trying to work through those details.

But thanks, and happy to accept other contributions. Also, feel free to leave thoughts/questions on issues that you're working on, happy to provide guidance on anything.

PR #767, opened 2017-01-15
fix: insert close-paran without breaking member access

refack:
I didn't bother with a bug. I just added a test (previously failing).
[eval](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=Q%20-%3E%20P%20-%3E%0D%0A%20%20foo%0D%0A.then%20(bar)%20-%3E%0D%0A%20%20bar)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules
 dist
 npm-debug.log
+.idea/
--- a/src/patchers/NodePatcher.js
+++ b/src/patchers/NodePatcher.js
@@ -1168,4 +1168,12 @@ export default class NodePatcher {
       receiver = receiver.parent;
     }
   }
+
+  /**
+   * Gets the first "interesting token" in the indexed range (default range is `this` + parent)
+   */
+  getFirstSemanticToken(from: number=this.contentStart, to: number=this.parent.contentEnd): ?SourceToken {
+    let nextSemanticIdx = this.indexOfSourceTokenBetweenSourceIndicesMatching(from, to, isSemanticToken);
+    return nextSemanticIdx && this.sourceTokenAtIndex(nextSemanticIdx);
+  }
 }
--- a/src/stages/normalize/patchers/FunctionApplicationPatcher.js
+++ b/src/stages/normalize/patchers/FunctionApplicationPatcher.js
@@ -82,7 +82,13 @@ export default class FunctionApplicationPatcher extends NodePatcher {
     let { args } = this;
     let lastArg = args[args.length - 1];
     if (lastArg.isMultiline()) {
-      this.insert(this.contentEnd, `\n${this.getIndent()})`);
+      // The CoffeeScript compiler will sometimes reject `.` that is starting a new line following a `)` token
+      let nextSemanticToken = this.getFirstSemanticToken(this.contentEnd);
+      if (nextSemanticToken && nextSemanticToken.type === SourceType.DOT) {
+        this.insert(nextSemanticToken.start, ')');
+      } else {
+        this.insert(this.contentEnd, `\n${this.getIndent()})`);
+      }
       return;
     }
 
--- a/test/member_access_test.js
+++ b/test/member_access_test.js
@@ -92,4 +92,49 @@ describe('member access', () => {
       a[b ? c : d];
     `);
   });
+
+  it('handles access to object constructed with multiline args 1', () => {
+    check(`
+      P -> 
+        foo
+      .then
+    `, `
+      P(() => foo).then;
+    `);
+  });
+
+  it('handles access to object constructed with multiline args 2', () => {
+    check(`
+      Q -> P -> 
+        foo
+      .then
+    `, `
+      Q(() => P(() => foo).then
+       );
+    `);
+  });
+
+  it('handles access to object constructed with multiline args 3', () => {
+    check(`
+      Q -> P -> 
+        foo
+        
+      .then
+    `, `
+      Q(() => P(() => foo).then
+       );
+    `);
+  });
+
+  it('handles access to object constructed with multiline args 4', () => {
+    check(`
+      Q -> P ->
+          foo
+          
+         .then
+    `, `
+      Q(() => P(() => foo).then
+       );
+    `);
+  });
 });


test/indentation_test.js
```diff
@@ -52,4 +52,16 @@ describe('indentation', () => {
       });
     `);
   });
+
+  it('handles indented lambda as arg to inline lambda', () => {
+    check(`
+    Q -> P ->
+      foo
+    .then (bar) ->
```
alangpierce: It looks like if there's another blank line after the `foo` line, it's still valid CoffeeScript but fails with the previous error. Can you try to make your change general enough to handle that?

src/stages/normalize/patchers/FunctionApplicationPatcher.js
```diff
@@ -82,7 +82,11 @@ export default class FunctionApplicationPatcher extends NodePatcher {
     let { args } = this;
     let lastArg = args[args.length - 1];
     if (lastArg.isMultiline()) {
-      this.insert(this.contentEnd, `\n${this.getIndent()})`);
+      if (this.slice(this.contentEnd, this.contentEnd + 2).match(/\n\s*\./)) {
```
alangpierce: The `+ 2` and the use of regexes seems a bit fragile. Maybe you could use one of the various `NodePatcher` to  methods to find a DOT token after the end of the expression?

test/indentation_test.js
```diff
@@ -52,4 +52,16 @@ describe('indentation', () => {
       });
     `);
   });
+
+  it('handles indented lambda as arg to inline lambda', () => {
+    check(`
+    Q -> P ->
```
alangpierce: Sample code should be indented one more level, so one indentation level past the `check`.

refack:
(I'll put this in the code)
I chose this codepoint because this is where the bug is and there's a need to be more careful when adding tokens. (like you mention in [line 57](https://github.com/decaffeinate/decaffeinate/pull/767/files/3f15c5427fa5f755f1ab35b942b3731c2bafe903#diff-12d668f38aed6588dd42a3d6d859d5a1R57))
This is a very weird case, I couldn't even mind-parse it correctly, as the DOT is assosiated with the `P` and not the `Q` or the `foo`... I couldn't find any way to fix it with indentation either.

I (manually) tested with other operators, but they all generate very different ASTs.

I would have liked to fix a more general case where we can actually convert single-statement-block-functions into lambdas, but I believe that should conceptually go in the `EsnextStage`.

refack:
Updated PR.
I figured it's actually an "member access" case so I moved the tests

.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules
 dist
 npm-debug.log
+.idea/
```
alangpierce: 👍 I use WebStorm as well. FYI you can always add this line to your local `.git/info/exclude` file to ignore it just for you, which is what I typically do when working with an open source project. But certainly no harm in adding this line to the .gitignore.

src/stages/normalize/patchers/FunctionApplicationPatcher.js
```diff
@@ -82,7 +82,17 @@ export default class FunctionApplicationPatcher extends NodePatcher {
     let { args } = this;
     let lastArg = args[args.length - 1];
     if (lastArg.isMultiline()) {
-      this.insert(this.contentEnd, `\n${this.getIndent()})`);
+      /**
+       * in case the next non-whitespace token is `.` we have to put the ) on it's line, so we lookahead.
```
alangpierce: For comments inside functions, use `//` instead of `/*` for comments. Also, comments should start with a capital letter.

Also, maybe mention that we need to do this because the CoffeeScript compiler will sometimes reject `)` followed by `.` on a later line.

src/stages/normalize/patchers/FunctionApplicationPatcher.js
```diff
@@ -82,7 +82,17 @@ export default class FunctionApplicationPatcher extends NodePatcher {
...
+       */
+      let nextInterestingIdx = this.indexOfSourceTokenBetweenSourceIndicesMatching(
+        this.contentEnd, this.parent.contentEnd, t => !this.slice(t.start, t.end).match(/^\s*$/));
```
alangpierce: I think the `isSemanticToken` method should handle that regex code (and it's often used by default). I think you may be able to do something similar to `getFollowingOpenParenToken` in #773 that I just put up.

.gitignore
refack: I know, but for my projects I actualy add the `.idea` folder and just exclude `.idea/workspace.xml`

refack: Ohh didn't read the whole thing (thought you meant the global .gitignore), got'ya.

refack:
I thought a more generalized `NodePatcher.prototype.getFirstSemanticToken` might come in handy in the future.

src/patchers/NodePatcher.js
```diff
@@ -1168,4 +1168,12 @@ export default class NodePatcher {
       receiver = receiver.parent;
     }
   }
+
+  /**
+   * Gets the first "interesting token" in the indexed range (default range is `this` + parent)
+   */
+  getFirstSemanticToken(from: number=this.contentStart, to: number=this.parent.contentEnd): ?SourceToken {
```
alangpierce: I imagine the default end of the range may need to be tweaked in the future if this gets used more, since I think sometimes a node can have the same end as its parent. And it might be best to just have no end and just find the next token. But looks like this works fine at least for this case.

refack: I wasn't even sure that there is always a parent, but `indexOfSourceTokenBetweenSourceIndicesMatching` requires a left end

alangpierce: Yep, everything has a parent except `ProgramPatcher`, which is rarely used directly. There should be a way to do an unbounded search, at the very least by using methods on the token list directly ( https://github.com/decaffeinate/coffee-lex/blob/master/src/SourceTokenList.ts ). But yeah, the token stuff is a little clunky sometimes.

refack:
I forgot how much I like compilation (only CS course  I got a 💯, and the funny thing is we build a compiler for a toy language called "Turkish Coffee")

PR #799, opened 2017-02-05
feat: handle class expressions for complex classes

### LINKED ISSUE #544: Cannot read property 'node' of null

```coffee
rethinkdb.monday = new (class extends RDBConstant then tt: protoTermType.MONDAY, st: 'monday')()
```

[(repl)](http://decaffeinate-project.org/repl/#?evaluate=false&stage=full&code=rethinkdb.monday%20%3D%20new%20(class%20extends%20RDBConstant%20then%20tt%3A%20protoTermType.MONDAY%2C%20st%3A%20'monday')())

I get this error:

```
Cannot read property 'node' of null
> 1 | rethinkdb.monday = new (class extends RDBConstant then tt: protoTermType.MONDAY, st: 'monday')()
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### LINKED ISSUE #625: Incorrect compiling coffeescript classes with prototype properties

```coffee
define (require) ->
  'use strict'


  class MainLayout extends BaseView
    container: 'body'
```

[Repl link](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=define%20(require)%20-%3E%0A%20%20'use%20strict'%0A%0A%0A%20%20class%20MainLayout%20extends%20BaseView%0A%20%20%20%20container%3A%20'body'%0A)

I get this output:

```js
define(function(require) {
  'use strict';


  class MainLayout extends BaseView {
    static initClass() {
      this.prototype.container = 'body';
    }
  }
  return MainLayout.initClass();
});
```

```js
define(function(require) {
  'use strict';


  class MainLayout extends BaseView {
    static initClass() {
      this.prototype.container = 'body';
    }
  }
  MainLayout.initClass();
  return MainLayout;
});
```

So instead of 

```js
return MainLayout.initClass();
```

which will return `undefined` it will be correct to have 

```js
MainLayout.initClass();
return MainLayout;
```

alangpierce:
Fixes #625

The general strategy here is to define an `__initClass__` helper in the
normalize stage which calls the `initClass` static method. This is uglier than
the previous approach of adding an `initClass` statement after the class
definition, so we dynamically decide which of the two to use, preferring the
statement approach. Since the normalize stage does not properly communicate
whether patching is done as an expression, we just look at the AST and only
allow the `initClass` statement if we're a child of a block and not in an
implicit return position.

Also, the existing code for handling helpers was all for the main stage, so I
factored it into a `SharedProgramPatcher` that also works in the normalize
stage.

This almost fixes #544, but there's a remaining issue in that example because
commas can split class properties but not statements in CoffeeScript.

There's still a little more to do to make complex classes 100% correct, but I
think this should handle most cases.

### DIFF
--- a/src/stages/normalize/patchers/ClassPatcher.js
+++ b/src/stages/normalize/patchers/ClassPatcher.js
@@ -8,9 +8,10 @@ import ProgramPatcher from './ProgramPatcher';
 import type { PatcherContext } from './../../../patchers/types';
 
 const INIT_CLASS_HELPER = `\
-__initClass__ = (c) ->
-  c.initClass()
-  return c
+\`function __initClass__(c) {
+  c.initClass();
+  return c;
+}\`
 `;
 
 export default class ClassPatcher extends NodePatcher {
--- a/src/stages/normalize/patchers/ClassPatcher.js
+++ b/src/stages/normalize/patchers/ClassPatcher.js
@@ -2,8 +2,16 @@ import { SourceType } from 'coffee-lex';
 
 import NodePatcher from './../../../patchers/NodePatcher';
 import AssignOpPatcher from './AssignOpPatcher';
+import BlockPatcher from './BlockPatcher';
+import ProgramPatcher from './ProgramPatcher';
 
 import type { PatcherContext } from './../../../patchers/types';
+
+const INIT_CLASS_HELPER = `\
+__initClass__ = (c) ->
+  c.initClass()
+  return c
+`;
 
 export default class ClassPatcher extends NodePatcher {
   nameAssignee: ?NodePatcher;
@@ -22,8 +30,6 @@ export default class ClassPatcher extends NodePatcher {
    * method instead.
    *
    * Current limitations:
-   * - Doesn't handle anonymous classes.
-   * - Doesn't handle classes used in an expression context.
    * - Doesn't deconflict the "initClass" name of the static method.
    * - Doesn't deconflict the variable assignments that are moved outside the
    *   class body.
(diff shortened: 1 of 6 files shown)


src/stages/normalize/patchers/ClassPatcher.js
```diff
@@ -2,9 +2,17 @@ import { SourceType } from 'coffee-lex';
 
 import NodePatcher from './../../../patchers/NodePatcher';
 import AssignOpPatcher from './AssignOpPatcher';
+import BlockPatcher from './BlockPatcher';
+import ProgramPatcher from './ProgramPatcher';
 
 import type { PatcherContext } from './../../../patchers/types';
 
+const INIT_CLASS_HELPER = `\
+__initClass__ = (c) ->
```
eventualbuddha: Should we just use interpolated JS via backticks?

test/class_test.js
```diff
@@ -1131,4 +1131,65 @@ describe('classes', () => {
...
+    check(`
+      A = class
+        b: c
+    `, `
+      const A = __initClass__(class {
+        static initClass() {
+          this.prototype.b = c;
+        }
+      });
+      var __initClass__ = function(c) {
```
eventualbuddha: This will fail since this helper won't be set yet. If we use the backtick approach for the helper we can use a function declaration which will be hoisted.

src/stages/normalize/patchers/ClassPatcher.js
alangpierce: Ooh, good idea. Still feels like sort of a hack (e.g. it could potentially mess up static analysis like name deconfliction in the main stage), but seems better than what I have, and of course it fixes the correctness issue you also mentioned.

alangpierce:
Ok, updated.

PR #808, opened 2017-02-12
fix: update decaffeinate-parser to latest

alangpierce:
There's now a `BareSuperFunctionApplication` node type that simplifies the
checks we needed to do. As before, it gets turned into its expanded form in the
normalize stage.

This also pulls in a fix to chained comparisons and unary operators.

### DIFF
--- a/src/stages/normalize/patchers/BareSuperFunctionApplicationPatcher.js
+++ b/src/stages/normalize/patchers/BareSuperFunctionApplicationPatcher.js
@@ -1,0 +1,7 @@
+import NodePatcher from './../../../patchers/NodePatcher';
+
+export default class FunctionApplicationPatcher extends NodePatcher {
+  patchAsExpression() {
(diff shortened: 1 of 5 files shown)


src/stages/normalize/patchers/BareSuperFunctionApplicationPatcher.js
```diff
@@ -0,0 +1,7 @@
+import NodePatcher from './../../../patchers/NodePatcher';
+
+export default class FunctionApplicationPatcher extends NodePatcher {
```
eventualbuddha: Change the name here.

alangpierce: Heh, oops, thanks.

alangpierce:
(Oh, oops, looks like that did a rebase when I wanted a squash.)

PR #815, opened 2017-02-13
fix: update decaffeinate-coffeescript and decaffeinate-parser to latest

alangpierce:
I think this should fix a build system failure that was happening earlier.

(Mark as fix to force a publish to update the repl.)


alangpierce:
Hmm, it passes locally for me. From the error, looks like it's using an old decaffeinate-coffeescript, but not sure how that could happen. Investigating further...

alangpierce:
Conclusion: specifying `^1.10.0-patch19` doesn't cause `npm install` to update to latest, which I guess is understandable since the semver rules get weird here. So I guess we should use `1.10.0-patch19` directly moving forward.

PR #831, opened 2017-02-18
refactor: add a `mayBeUnboundReference` to `NodePatcher`

alangpierce:
This makes it easier to determine when we need to do a `typeof` check, and
will be useful for some upcoming changes to soak operations that will need
to do a similar check.

### DIFF
--- a/src/patchers/NodePatcher.js
+++ b/src/patchers/NodePatcher.js
@@ -1178,7 +1178,7 @@ export default class NodePatcher {
    * value, to guard against the case where this node is a variable that doesn't
    * exist. IdentifierPatcher overrides this to check the current scope.
    */
-  mayBeInvalidReference() {
+  mayBeUnboundReference() {
     return false;
   }
 }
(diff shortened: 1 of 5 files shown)


src/patchers/NodePatcher.js
```diff
@@ -1172,4 +1172,13 @@ export default class NodePatcher {
     let nextSemanticIdx = this.indexOfSourceTokenBetweenSourceIndicesMatching(from, to, isSemanticToken);
     return nextSemanticIdx && this.sourceTokenAtIndex(nextSemanticIdx);
   }
+
+  /**
+   * Determine if we need to do a `typeof` check in a conditional for this
+   * value, to guard against the case where this node is a variable that doesn't
+   * exist. IdentifierPatcher overrides this to check the current scope.
+   */
+  mayBeInvalidReference() {
```
eventualbuddha: I might have named this `mayBeUnboundReference`. What do you think?

alangpierce: Sure, that seems more clear.

PR #854, opened 2017-02-20
fix: don't specify a class name for namespaced classes

### LINKED ISSUE #853: Crash on class this-assigned to a keyword name

```coffee
class @for
```

[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=class%20%40for)

I get this error:

```
AddVariableDeclarationsStage failed to parse: Unexpected token, expected { (1:17)
> 1 | this.for = class for {};
    |                  ^
```

I think it's actually incorrect to try to put the name of the class on the RHS anyway, since in normal conditions it introduces a new variable name when it shouldn't.

alangpierce:
Fixes #853

The code to put a name for the class was actually wrong, since it introduces a
name when it shouldn't. And in this case, it also introduced the use of a
keyword as an identifier, which caused a crash.

### DIFF
--- a/src/stages/main/patchers/ClassPatcher.js
+++ b/src/stages/main/patchers/ClassPatcher.js
@@ -1,6 +1,5 @@
 import ClassBlockPatcher from './ClassBlockPatcher';
 import IdentifierPatcher from './IdentifierPatcher';
-import MemberAccessOpPatcher from './MemberAccessOpPatcher';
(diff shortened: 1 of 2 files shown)


alangpierce:
@eventualbuddha oops, you're right, I misinterpreted it. I'll file a bug and fix when I get a chance. Rough fix will be to revert this change but also put in a special case for when the resolved name is a keyword.

PR #859, opened 2017-02-20
fix: allow duplicate class declarations

### LINKED ISSUE #855: Declaring the same class name twice causes a crash

```coffee
class A
class A
```

[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=class%20A%0Aclass%20A)

I get this error:

```
Cannot read property 'file' of undefined
```

In CoffeeScript, class statements just cause an assignment, while in JavaScript, they count as a declaration, so you can't have two with the same name.

alangpierce:
Fixes #855

We now track class declarations in our scope, and only use the normal class
syntax if the class being patched is also the declaration site (the first
assignment seen by the scope).

This also lets us remove a special case in the `initClass` patching in the
normalize step. We don't need to explicitly create an assignment anymore,
because one is created automatically due to a variable with the same name being
declared above the class declaration.

### DIFF
--- a/test/class_test.js
+++ b/test/class_test.js
@@ -881,7 +881,7 @@ describe('classes', () => {
       let C = undefined;
       class A {
         static initClass() {
-          B = class B {
+          B = class {
             static initClass() {
               this.prototype.classField = 2;
             }
@@ -891,7 +891,7 @@ describe('classes', () => {
           };
           B.initClass();
           let CONSTANT = undefined;
-          C = class C {
+          C = class {
             static initClass() {
               CONSTANT = 7;
             }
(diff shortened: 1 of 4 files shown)


test/class_test.js
```diff
@@ -1305,4 +1305,14 @@ describe('classes', () => {
...
     `);
   });
+
+  it('allows two class declarations with the same name', () => {
+    check(`
+      class A
+      class A
+    `, `
+      class A {}
+      A = class {};
```
eventualbuddha: Wouldn't we want to keep the name on the class expression? It could matter.

PR #894, opened 2017-03-04
fix: don't crash on files with mixed tabs and spaces

### LINKED ISSUE #893: Crash on file with mixed indentation

```coffee
while a
  b

while c
	if d
		e
```

[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=while%20a%0A%20%20b%0A%0Awhile%20c%0A%09if%20d%0A%09%09e)

I get this error:

```
cannot unindent line by -1 without enough indent
  1 | while a
> 2 |   b
    |  ^
  3 | 
  4 | while c
```

From https://github.com/decaffeinate-examples/codecombat/blob/decaffeinate/scripts/minify.coffee

alangpierce:
Fixes #893

Rather than crashing when we run into an unexpected indentation issues, we
should just silently ignore them for now. In the test case, the generated code
ended up being fine anyway.

### DIFF
--- a/src/patchers/NodePatcher.js
+++ b/src/patchers/NodePatcher.js
@@ -1084,11 +1084,16 @@ export default class NodePatcher {
             } else if (source.slice(i - indentToChange.length, i) === indentToChange) {
               this.remove(i - indentToChange.length, i);
             } else {
-              throw this.error(
-                `cannot unindent line by ${offset} without enough indent`,
-                i - indentToChange.length,
-                i
-              );
+              // Ignore this case: we're trying to unindent a line that doesn't
+              // start with enough indentation, or doesn't start with the right
+              // type of indentation, e.g. it starts with spaces when the
+              // program indent string is a tab. This can happen when a file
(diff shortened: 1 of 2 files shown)


src/patchers/NodePatcher.js
```diff
@@ -1084,11 +1084,15 @@ export default class NodePatcher {
...
-              );
+              // Ignore this case: we're trying to unindent a line that doesn't
+              // start with enough indentation, or doesn't start with the right
+              // type of indentation, e.g. it starts with spaces when the
+              // program indent string is a tab. This can happen when a file
+              // uses inconsistent indentation in different parts. We only
+              // expect this to come up in the main stage, so getting
+              // indentation wrong means ugly JS code that's still correct.
+              // If there's ever a warning system, it would be good to log a
+              // warning here.
```
eventualbuddha: We could log it using `this.log`.

alangpierce: Good point, I'll add that.

PR #903, opened 2017-03-06
fix: don't unsafely negate operators like `>` by default

### LINKED ISSUE #902: Comparison operator negation rules can be incorrect

```coffee
unless NaN > 1
  console.log 'first'
else
  console.log 'second'
```

[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=unless%20NaN%20%3E%201%0A%20%20console.log%20'first'%0Aelse%0A%20%20console.log%20'second')

I get this output:

```js
if (NaN <= 1) {
  console.log('first');
} else {
  console.log('second');
}
```

```js
if (!(NaN > 1)) {
  console.log('first');
} else {
  console.log('second');
}
```

This is causing at least two problems in atom. I think this negation behavior probably needs to be behind a `--loose` flag, since it doesn't work for operands like `NaN` and `undefined`.

alangpierce:
Fixes #902

For both normal binary operators and chained comparison operators, we need to be
careful about rewriting `!(a < b)` as `a >= b`, since the rewrite isn't correct
for cases like `NaN`. Now, we just do the simpler negation by default and only
do the nicer-looking one with an opt-in flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ problems you run into.
 * `--loose-for-expressions`: Do not wrap expression loop targets in `Array.from`.
 * `--loose-for-of`: Do not wrap JS `for...of` loop targets in `Array.from`.
 * `--loose-includes`: Do not wrap in `Array.from` when converting `in` to `includes`.
+* `--loose-comparison-negation`: Allow unsafe simplifications like `!(a > b)` to `a <= b`.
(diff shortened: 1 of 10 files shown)


alangpierce:
I think the switch negation was using normal comparison operator negation. Do you know more specifically what case might have problems. Regardless, I'll merge now so it makes it into today's build.

PR #905, opened 2017-03-07
fix: ensure that inclusive slice bounds are numbers

### LINKED ISSUE #882: Inclusive slice operation with string end gives wrong result

```coffee
a = [1, 2, 3, 4, 5]
start = '1'
end = '3'
console.log a[start..end]
```

[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&stage=MainStage&code=a%20%3D%20%5B1%2C%202%2C%203%2C%204%2C%205%5D%0Astart%20%3D%20'1'%0Aend%20%3D%20'3'%0Aconsole.log%20a%5Bstart..end%5D%0A)

I get this output:

```js
a = [1, 2, 3, 4, 5];
start = '1';
end = '3';
console.log(a.slice(start, end + 1 || undefined));
```

```js
a = [1, 2, 3, 4, 5];
start = '1';
end = '3';
console.log(a.slice(start, +end + 1 || undefined));
```

Otherwise, it takes the range from '1' to '31'.

alangpierce:
Fixes #882

We need to put a `+` unary op at the front to avoid string concatenation when
adding `1`.

### DIFF
--- a/src/stages/main/patchers/SlicePatcher.js
+++ b/src/stages/main/patchers/SlicePatcher.js
@@ -62,9 +62,9 @@ export default class SlicePatcher extends NodePatcher {
             `, ${right.node.data + 1}`
           );
         } else {
-          // `a.slice(0..1]` → `a.slice(0, 1]`
-          //           ^^                ^^
-          this.overwrite(slice.start, slice.end, ', ');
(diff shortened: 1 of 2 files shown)


alangpierce:
Good catch, thanks.

PR #1072, opened 2017-06-12
fix: extract loop variables to their own name when necessary

### LINKED ISSUE #1044: for-in loops with indices allow modifying the loop counter when they shouldn't

```coffee
arr = [1, 2, 3]
for val, i in arr
  console.log i
  i = 10
```

[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=arr%20%3D%20%5B1%2C%202%2C%203%5D%0Afor%20val%2C%20i%20in%20arr%0A%20%20console.log%20i%0A%20%20i%20%3D%2010)

I get this output:

```js
let arr = [1, 2, 3];
for (let i = 0; i < arr.length; i++) {
  let val = arr[i];
  console.log(i);
  i = 10;
}
```

Here's what I expect it to be instead (one idea):

```js
let arr = [1, 2, 3];
for (let j = 0; j < arr.length; j++) {
  let i = j;
  let val = arr[i];
  console.log(i);
  i = 10;
}
```

alangpierce:
Fixes #1044

We need to do this when the loop body might assign to the loop variable. We can
detect this by determining if there are any assignments to the variable directly
in the loop body or if there are any instances of the variabble assigned from a
closure anywhere in the function scope.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "automatic-semicolon-insertion": "^1.0.2",
     "babylon": "^6.12.0",
     "coffee-lex": "^7.0.0",
-    "decaffeinate-coffeescript": "1.10.0-patch22",
+    "decaffeinate-coffeescript": "1.10.0-patch24",
     "decaffeinate-parser": "^17.1.3",
     "detect-indent": "^4.0.0",
     "esnext": "^3.1.0",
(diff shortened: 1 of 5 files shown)


eventualbuddha:
This turned out to be complicated! Thanks 😊

PR #1126, opened 2017-07-06
refactor: convert various utils files to typescript

alangpierce:
Also upgrade various relevant dependencies.


src/utils/formatCoffeeScriptAst.ts
```diff
@@ -70,7 +72,7 @@ function shouldTraverse(value) {
  * CoffeeScript AST nodes are always instances of a custom class, so use the
  * constructor name to distinguish between node children and non-node children.
  */
-function isNode(value) {
+function isNode(value: {}): boolean {
```
eventualbuddha: Eventually this should be replaced with a better check.

src/utils/formatCoffeeScriptLocationData.ts
```diff
@@ -0,0 +1,16 @@
+import { LocationData } from 'decaffeinate-coffeescript/lib/coffee-script/nodes';
+import ParseContext from 'decaffeinate-parser/dist/util/ParseContext';
+import formatRange from './formatRange';
+
+export default function formatCoffeeScriptLocationData(locationData: LocationData, context: ParseContext) {
```
eventualbuddha: Annotate return value? `: string`

src/utils/formatRange.ts
```diff
@@ -7,14 +7,20 @@
...
  */
-export default function formatRange(startIndex: number, endIndex: number, context) {
+import ParseContext from 'decaffeinate-parser/dist/util/ParseContext';
+
+export default function formatRange(startIndex: number, endIndex: number, context: ParseContext) {
   return `[${formatIndex(startIndex, context)}-${formatIndex(endIndex, context)}]`;
 }
 
-function formatIndex(index: number, context) {
+function formatIndex(index: number, context: ParseContext) {
```
eventualbuddha: Also annotate return value.

src/utils/requiresParentheses.js
```diff
@@ -1,30 +0,0 @@
-/* @flow */
```
eventualbuddha: I'm guessing this file was unused?

alangpierce: Yep.

src/utils/formatCoffeeScriptAst.ts
alangpierce: I'm not sure there's really a better type here, although maybe `any` and silencing the lint rule would be better. This is reflectively walking CoffeeScript AST nodes to show in the repl, and doing some runtime heuristics to figure out what is and isn't a node so it knows when to traverse.

eventualbuddha: Perhaps `any` would be good. I think the right return type may actually be `value is Node` rather than `boolean`, as that provides TypeScript with the ability to refine types using `isNode` calls. For example, it would know that in this `if` block that `n` is a `Node`:

```ts
If (isNode(n)) {
  // do something with `n` that assumes it’s a `Node`
}
```

eventualbuddha: Then again… why doesn’t `instanceof Node` work?

alangpierce: Oh, yeah, you're right, `instanceof Base` works just fine. My line of reasoning was "it's not exported by CoffeeScript and types are erased, so there's no way TypeScript can generate code that works". But it actually is exported by CoffeeScript, just not at the top level. Will fix in a follow-up commit. (But for `shouldTraverse`, I think `any` is probably still reasonable.)

And thanks for pointing me to the `value is Node` syntax (user-defined type guards). I had no idea TypeScript had that.

PR #1131, opened 2017-07-07
Add instruction to convert all files in folder/project

hneiva:


alangpierce:
Hmm, I think passing a directory to decaffeinate is supposed to make it recursively walk the directory and convert all CS files, but that seems to not be happening when I test it (it only goes one level deep). Looks like probably a bug; `decaffeinate .` should work. I'll take a closer look when I get a chance.

hneiva:
That makes sense, I tried here like you suggested and same happened. If I have time later, I'll dig into the code and find out why it's not working

alangpierce:
Sure, that would be great. It's `runWithPaths` in cli.js: https://github.com/decaffeinate/decaffeinate/blob/master/src/cli.js#L143

The code was written quite a while ago, and it's possible that the original intention was to not do a recursive traversal, but it looks like it tries to do that, and doing a recursive traversal makes the most sense to me.

Also, FWIW, I think [bulk-decaffeinate](https://github.com/decaffeinate/bulk-decaffeinate) (or a similar tool) is generally better for converting a directory or codebase all at once, and I'd want that to be fairly prominent in the docs (although I understand it's a bit simpler to just run decaffeinate directly).

eventualbuddha:
Looks like it’s because of [this filter](https://github.com/decaffeinate/decaffeinate/blob/master/src/cli.js#L164) which only passes things through that have specific file extensions. We want to process directories too, but they’re filtered out unless they happen to have one of the looked-for extensions.

eventualbuddha:
When this is fixed it’d be a good idea to add a test verifying it.

alangpierce:
👍 there are already tests for file discovery at the bottom of `cli_test.js` and using the `test/fixtures` directory, so I think a test for this should be pretty easy.

hneiva:
Updated the code, let me know what you guys think

alangpierce:
@hneiva thanks! Would you mind rebasing? I just finished converting everything to TypeScript and moved the fixtures directory to top-level test_fixtures instead of test/fixtures. I think the rebase should be reasonably straightforward, but sorry for the conflicts!

hneiva:
Now why did it pull so many changes at once.
I think I dun goof'd?

src/cli.ts
```diff
@@ -157,23 +157,32 @@ function runWithPaths(paths: Array<string>, options: CLIOptions, callback: ((err
...
+            info.isDirectory() ||
+            (options.modernizeJS && child.endsWith('.js')) ||
+            // Extensions .coffee || .litcoffee || .coffee.md
+            /\.(lit)?coffee(\.md)?$/.test(child)
+          ) {
+            pending.push(join(path, child));
+          }
+
+          // On last run, process next
+          if (children.length === index + 1) {
```
alangpierce: Hmm I think this actually has a race condition. All of the stat calls are happening in parallel and might come back in any order, so I think the very last one (which might be non-coffeescript) could come back first, call `processNext` before the others get added, and none of the other files will get processed.

alangpierce:
Actually, nevermind, I think my suggested approach would break the "properly converts an unrecognized extension" test; we want to filter the file extensions when they're auto-discovered, but not when they're explicitly specified. Another idea: rather than doing `children.length === index + 1`, you could keep a counter and call `processNext` when the last `stat` call finishes.

hneiva:
@alangpierce ok, I've updated the CLI to correctly try all files (except a few)
Not completely sure what files we should be ignoring, I've excluded `.gitignore`, `node_modules` and any files that starts with `.` (hidden/system files)
Maybe we shouldn't assume this? What do you think?

hneiva:
Hey @alangpierce, I was messing around with `async`, but I'm not sure why it won't work with the cli command. Mind having a look to see if you can figure it out?
https://github.com/hneiva/decaffeinate/blob/async-cli/src/cli.ts

alangpierce:
@hneiva I looked at your code and it looks like it's failing because the `children.length === index + 1` check never happens due to the last item being a directory (you `return` early in the directory case).

More generally, you probably shouldn't mix callbacks, promises, and async functions like that. For example, you're still running the callback-based `stat` within the `forEach`, which is creating a bunch of promises in parallel. With async functions you can just do `await` within a loop to make sure things are done in series. It's also probably best to use the `mz` package, which provides promise equivalents of all of the node functions. Ideally, async functions are used for all of the control flow, and `new Promise` is only used in isolated situations to bridge with old code expecting callbacks.

bulk-decaffeinate uses async functions to do a really similar thing, so you could use that for inspiration:
https://github.com/decaffeinate/bulk-decaffeinate/blob/master/src/util/getFilesUnderPath.js

(I'm also happy to take this on if you want; either way. I've been somewhat meaning to switch to async functions here, and I think it'll make the code a lot more straightforward.)

eventualbuddha:
+1 to async functions, probably using `mz/fs`.

hneiva:
Sorry guys, had to deal with some fires at work this week (still going, but less crazy). 
And the forest fires around my place in BC aren't helping either..
I'll try and get to it sometime this weekend.

src/cli.ts
```diff
@@ -1,9 +1,9 @@
 /* eslint-disable no-process-exit */
 
-import { readdir, readFile, stat, writeFile } from 'fs';
-import { basename, dirname, extname, join } from 'path';
-import { convert, modernizeJS } from './index';
-import { Options } from './options';
+import {readdir, readFile, stat, writeFile} from 'mz/fs';
+import {basename, dirname, extname, join} from 'path';
+import {convert, modernizeJS} from './index';
+import {Options} from './options';
```
eventualbuddha: In a separate PR we should enforce the spacing convention via tslint. In the meantime, can you change these back and match the existing style?

src/cli.ts
```diff
@@ -144,43 +144,54 @@ function runWithPaths(paths: Array<string>, options: CLIOptions, callback: ((err
...
+    const info = await stat(path);
+
+    if (!info) {
+      // Skip
+      return processNext();
+    }
+
+    if (info.isDirectory()) {
+      return processDirectory(path)
+        .then(processNext);
```
eventualbuddha: Let's not mix manual promise syntax with a sync/await. This should be:

```ts
await processDirectory(path);
await processNext();
```

And probably just have an `else` containing `await processFile(path);`.

src/cli.ts
```diff
@@ -144,43 +144,54 @@ function runWithPaths(paths: Array<string>, options: CLIOptions, callback: ((err
...
+      }
+    }
+  }
+
+  function processDirectory(path: string): Promise<void> {
+    return readdir(path)
+      .then(children => processDirectoryChildren(children, path))
+      .catch(err => {
+        errors.push(err);
+      });
```
eventualbuddha: We can also make this use async/await. Just think of it as synchronous code:

```ts
try {
  let children = await readdir(path);
  await processDirectoryChildren(children, path);
} catch (err) {
  errors.push(err);
}

File: src/cli.ts [hunk 5]
```diff
@@ -144,43 +144,56 @@ function runWithPaths(paths: Array<string>, options: CLIOptions, callback: ((err
...
-        processDirectory(path);
-      } else {
-        processFile(path);
+  async function processPath(path: string): Promise<void> {
+    try {
+      const info = await stat(path);
+
+      if (info.isDirectory()) {
+        await processDirectory(path);
+        return processNext();
```
eventualbuddha: I'd make this `await processNext();` since we don't actually want to propagate any return value.

hneiva:
I modified a bit how `processFile` works (doesn't call `processNext`). Let me know if you guys think `processNext` should only be called after `processFile` has finished.

src/cli.ts
```diff
@@ -134,7 +134,7 @@ function parseArguments(args: Array<string>): CLIOptions {
     }
   }
 
-  return { paths, baseOptions, modernizeJS };
```
eventualbuddha: Unnecessary white space change.

hneiva: I honestly think WebStorm should die in a fire sometimes.

src/cli.ts
```diff
@@ -144,67 +144,81 @@ function runWithPaths(paths: Array<string>, options: CLIOptions, callback: ((err
...
-        );
+  async function processDirectoryChildren(children: Array<string>, path: string): Promise<void> {
+    for (let child of children) {
+      try {
+        const childPath = join(path, child);
+        const info = await stat(childPath);
+
+        if (info.isDirectory()) {
+          await processDirectory(childPath);
+          continue;
```
eventualbuddha: Why `continue` rather than an `else if` clause for the file case?

hneiva: Personal preference of using guard statements over if elses.

eventualbuddha:
@hneiva @alangpierce what do you think of the updates I made?

src/cli.ts
```diff
@@ -151,74 +151,73 @@ function parseArguments(args: Array<string>): CLIOptions {
...
-      let nextPath = pending.shift();
-      if (!nextPath) {
-        throw new Error('Expected a next path.');
-      }
-      processPath(nextPath);
-    } else if (callback) {
-      callback(errors);
+    try {
+      let data = await readFile(path, 'utf8');
+      let resultCode = runWithCode(path, data, options);
```
alangpierce: Putting this in a try/catch actually causes decaffeinate to swallow all unexpected errors now. If an error comes up that's not a PatchError, it throws here, gets added to `errors`, and then gets ignored since `errors` and the `callback` are never used.

eventualbuddha:
And, finally done. Thanks, @hneiva!

PR #1134, opened 2017-07-08
fix: convert all normalize stage patchers to TypeScript

alangpierce:
Some of these could be done independently, but there were about 7 files that had
references between each other. Generally these were straightforward, although I
reworked the control flow a little in a few places, and I had to make
NodePatcher aware of the "tack a callback on so a child can use it" approach
that FunctionPatcher does.

Mark as fix to force a release so we don't go too long without running the full
test suite of projects.

This also involved changing the typescript settings to allow doing for-of loops
and spreads over iterables using the downlevelIteration option. This introduced
usages of new TypeScript helpers that caused a crash when running the binary. I
think the rollup typescript plugin was trying to add all helpers, but it had an
out-of-date list, so instead I configured TypeScript to import from tslib
instead, then I added it as a dependency.


package.json
```diff
@@ -46,7 +46,8 @@
     "detect-indent": "^4.0.0",
     "esnext": "^3.1.7",
     "lines-and-columns": "^1.1.5",
-    "magic-string": "^0.22.1"
+    "magic-string": "^0.22.1",
+    "tslib": "^1.7.1"
```
eventualbuddha: Do you think we should just do what we did for coffee-lex and decaffeinate-parser and just not use rollup?

alangpierce: Probably (although it's less important since we don't need typescript typings for decaffeinate I think). But I think the config changes here are a good idea regardless, since you get proper for-of semantics while also not generating redundant helpers.

I just tried it on coffee-lex where it's a little easier to see the JS output.

First I enabled `"downlevelIteration": true` and added a for loop over a string (which otherwise TS would complain about). This resulted in ~30 lines of helper functions added to the top of the file.

Then I set `"noEmitHelpers": true`, which got rid of the helper definitions and broke tests because the helpers weren't defined at the global scope.

Then I set `"importHelpers": true`, which worked, but requires `tslib` to be a dependency (not just a devDependency).

So I think all three of three of those settings plus the tslib dependency are generally a good idea for any project targeting ES5 or earlier.

But yeah, it's annoying that rollup breaks the helpers thing. I think they might try to do something special and they just have an out-of-date list of helpers? I needed `__values`, `__spread`, and `__read`.
https://github.com/rollup/rollup-plugin-typescript/blob/master/src/typescript-helpers.js

It does still seem nice to be able to use rollup in a TypeScript project, so a more ambitious option is to fix the plugin to make it work like we want.

eventualbuddha:
Yeah, see https://github.com/rollup/rollup-plugin-typescript/pull/91.

Issue #1201, opened 2017-11-04
Variable incorrectly hoisted for => functions.


samertm:
decaffeinate is producing the wrong JavaScript based on my CoffeeScript input:

```coffee
# a is incorrectly hoisted out of the => functions
() =>
  a = 1
  return

() =>
  a = 2
  return

# b is not hoisted out of -> functions
() ->
  b = 1
  return

() ->
  b = 2
  return
```

[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=%23%20a%20is%20incorrectly%20hoisted%20out%20of%20the%20%3D%3E%20functions%0A()%20%3D%3E%0A%20%20a%20%3D%201%0A%20%20return%0A%0A()%20%3D%3E%0A%20%20a%20%3D%202%0A%20%20return%0A%0A%23%20b%20is%20not%20hoisted%20out%20of%20-%3E%20functions%0A()%20-%3E%0A%20%20b%20%3D%201%0A%20%20return%0A%0A()%20-%3E%0A%20%20b%20%3D%202%0A%20%20return)

I get this output:

```js
// a is incorrectly hoisted out of the => functions
let a;
() => {
  a = 1;
};

() => {
  a = 2;
};

// b is not hoisted out of -> functions
(function() {
  const b = 1;
});

(function() {
  const b = 2;
});
```

Here's what I expect it to be instead:

```js
// a is incorrectly hoisted out of the => functions
(() => {
  var a;
  a = 1;
});

(() => {
  var a;
  a = 2;
});

// b is not hoisted out of -> functions
(function() {
  var b;
  b = 1;
});

(function() {
  var b;
  b = 2;
});
```

[(coffeescript repl)](http://coffeescript.org/#try:%23%20a%20is%20incorrectly%20hoisted%20out%20of%20the%20%3D%3E%20functions%0A()%20%3D%3E%0A%20%20a%20%3D%201%0A%20%20return%0A%0A()%20%3D%3E%0A%20%20a%20%3D%202%0A%20%20return%0A%0A%23%20b%20is%20not%20hoisted%20out%20of%20-%3E%20functions%0A()%20-%3E%0A%20%20b%20%3D%201%0A%20%20return%0A%0A()%20-%3E%0A%20%20b%20%3D%202%0A%20%20return)

alangpierce:
Thanks for reporting! Probably just a missing case in the add-variable-declarations step, hopefully should be an easy fix.

[...]

PR #1204, opened 2017-11-05
fix: don't generate blockless arrow functions that contain assignments

### LINKED ISSUE #1202: Assignment within blockless arrow function is incorrectly hoisted

```coffee
f = => a = 1
a = 2
f()
console.log a
```

[(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&stage=full&code=f%20%3D%20%3D%3E%20a%20%3D%201%0Aa%20%3D%202%0Af()%0Aconsole.log%20a)

I get this output:

```js
let a;
const f = () => a = 1;
a = 2;
f();
console.log(a);
```

It prints 1.

```js
const f = () => {
  let a;
  return a = 1;
}
const a = 2;
f();
console.log(a);
```

It should print 2.

Some additional discussion here: https://github.com/eventualbuddha/add-variable-declarations/pull/79#pullrequestreview-74259022

alangpierce:
Fixes #1202

add-variable-declarations won't have a place to add a declaration if the
function doesn't have a block body, so if there's any assignment in an arrow
function, we always give it a block body.

The formatting after decaffeinate and add-variable-declarations looks really
bad, I think because add-variable-declarations doesn't attempt to handle inline
functions, but this case is rare enough that I'm not too worried about
formatting.

Also update add-variable-declarations and add a regression test for #1201.

### DIFF
--- a/src/stages/main/patchers/BoundFunctionPatcher.ts
+++ b/src/stages/main/patchers/BoundFunctionPatcher.ts
@@ -1,7 +1,8 @@
-import { Node } from 'decaffeinate-parser/dist/nodes';
+import { AssignOp, Node } from 'decaffeinate-parser/dist/nodes';
 import { PatcherClass } from '../../../patchers/NodePatcher';
 import { PatchOptions } from '../../../patchers/types';
 import blockStartsWithObjectInitialiser from '../../../utils/blockStartsWithObjectInitialiser';
+import containsDescendant from '../../../utils/containsDescendant';
 import notNull from '../../../utils/notNull';
 import referencesArguments from '../../../utils/referencesArguments';
 import FunctionPatcher from './FunctionPatcher';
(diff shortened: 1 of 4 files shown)


samertm:
Wow, thanks so much for the quick fix! You all have done an awesome job with decaffeinate, btw! Do you think you could cut a release after this lands?

alangpierce:
@samertm 👍 Thanks!

The bug you filed (#1201) was already fixed with add-variable-declarations 3.1.2 ( https://github.com/eventualbuddha/add-variable-declarations/pull/79 ), which matches decaffeinate's dependency `"add-variable-declarations": "^3.1.0"`. So installing decaffeinate from scratch should pull that in.

decaffeinate and its sub-projects use [semantic-release](https://github.com/semantic-release/semantic-release), so a release will be generated automatically once this PR merges. The "fix:" at the start of the commit message indicates that it'll be a patch release, so in this case it'll be 3.3.12.

PR #1213, opened 2017-12-16
chore(package): move `@types` packages to `devDependencies`

eventualbuddha:


alangpierce:
Thanks! I guess I put those there by accident.

PR #1232, opened 2018-02-12
docs: document that inline JS should not rely on CS implementation details

alangpierce:


docs/correctness-issues.md
```diff
@@ -85,7 +109,7 @@ after decaffeinate when run in a modern JS implementation:
 new =>
 ```
 
-This affects "`new` works against bare function" test in the CoffeeScript test
+This affects the "`new` works against bare function" test in the CoffeeScript test
```
eventualbuddha: Intentional?

eventualbuddha: Ah, nevermind. I see it now.

PR #1235, opened 2018-02-12
docs(issues): use the new Github multiple issue templates feature

eventualbuddha:
We retain the default one for someone who just stumbles onto the issues page, but if you come from the README we can be a little more targeted.

https://github.com/blog/2495-multiple-issue-and-pull-request-templates

### DIFF
--- a/.github/ISSUE_TEMPLATE/crash.md
+++ b/.github/ISSUE_TEMPLATE/crash.md
@@ -1,0 +1,23 @@
+<!--
+
+☕️ 𝓭𝓮𝓬𝓪𝓯𝓯𝓮𝓲𝓷𝓪𝓽𝓮
+
+Thanks for creating an issue! You can paste your code into http://decaffeinate-project.org/repl/ and then copy the URL to get a shareable link to your code. Pasting that URL into the issue as a markdown link is helpful, like so:
+
+    [(repl)](http://decaffeinate-project.org/repl/#?evaluate=true&code=a%20b)
+
+-->
+
+decaffeinate is crashing on my CoffeeScript input:
+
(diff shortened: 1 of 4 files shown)


eventualbuddha:
It actually [is possible](https://github.com/decaffeinate/decaffeinate/issues/new?title=test&body=bar). Want to do that as a follow-up to this?

alangpierce:
Ooh neat, thanks for the pointer. I may eventually try to get that working for the decaffeinate repl, although hopefully there won't be too many more bugs anyway. But there's [another side project](https://sucrase.io/) I've been working on where I've thought about doing something like that.

eventualbuddha:
Nice side project 😊 I assume you’ve seen buble? I’m curious why you didn’t go with that.

alangpierce:
Thanks 😄. Yep, I've seen Buble, and had it in the performance comparison for a while. Buble compiles a subset of ES2015 (plus JSX) to ES5, so it doesn't support most code I want to write. Sucrase compiles JSX, TypeScript, Flow, imports, and a few JS proposals to be able to run in node 8 and latest Chrome, so that I get fast development builds. Disabling most Babel transforms doesn't speed Babel up very much, so this was sort of re-thinking the architecture to be much faster when only running a few (relatively simple) transforms.

Architecturally, Sucrase is a little more aggressive than Buble in that it tries to get away with only token-based transforms rather than parsing to an AST. (In practice, the Babylon fork ended up being somewhere between a parser and a lexer.) It also builds the resulting code on-the-fly rather than using magic-string. I'm hoping to get it to a point where it can run as WebAssembly, which hopefully would make it even faster.

PR #1248, opened 2018-03-05
Use async/await in CLI

eventualbuddha:
Looking at #1131 I decided to take another stab at doing async/await with the CLI functions. I opted to keep the `pending` queue and the specific functions in use to keep the refactoring simple. The biggest difference is that error handling is done at a higher level, in `processAll`.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -10,21 +10,21 @@ const pkg = require('../package');
 /**
  * Run the script with the user-supplied arguments.
  */
-export default function run(args: Array<string>): void {
+export default async function run(args: Array<string>): Promise<void> {
   let options = parseArguments(args);
 
   if (options.paths.length) {
-    runWithPaths(options.paths, options);
+    await runWithPaths(options.paths, options);
   } else {
-    runWithStdio(options);
+    await runWithStdio(options);
   }
 }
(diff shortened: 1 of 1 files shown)


src/cli.ts
```diff
@@ -10,21 +10,21 @@ const pkg = require('../package');
...
+    await runWithStdio(options);
   }
 }
 
-type CLIOptions = {
-  paths: Array<string>,
-  baseOptions: Options,
-  modernizeJS: boolean,
-};
+interface CLIOptions {
```
alangpierce: Was changing `type` to `interface` necessary here? I had viewed them as pretty much interchangeable, but looks to me like interface is actually preferred if possible (e.g. it gives better error messages), was that the rationale?

alangpierce:
> I opted to keep the pending queue
Oops, missed this part. Makes sense to keep the changes small.

src/cli.ts
eventualbuddha: I've seen hand-wavy things indicating that interface is usually better, including [this tslint rule](https://palantir.github.io/tslint/rules/interface-over-type-literal/), so I figured I'd just change it while I was there.

PR #1249, opened 2018-03-06
fix(cli): stop ignoring non-patch exceptions

eventualbuddha:
We no longer catch errors at this level, so non-patch exceptions will bubble up and likely crash the program rather than being silently swallowed. I consider this a good thing, hence the `fix` designation. However, it could be a breaking change for some use cases. Since I never intended it to ignore errors, I think releasing this as a patch is actually correct, but it's worth noting.

This cleans up a few things too:

- remove unused `callback` parameter
- remove `errors` array as we never used the result anyway
- just await the result of calling `processDirectory` rather than pushing onto the `pending` array

These are all things @alangpierce pointed out in #1248.


src/cli.ts
```diff
@@ -166,19 +163,18 @@ async function runWithPaths(paths: Array<string>, options: CLIOptions, callback:
...
-          } else {
-            return child.endsWith('.coffee') ||
-              child.endsWith('.litcoffee') ||
-              child.endsWith('.coffee.md');
-          }
-        })
-        .map(child => join(path, child))
-    );
+
+    for (let child of children) {
```
alangpierce: Looks like this isn't traversing subdirectories, as mentioned in #1131. With this refactor, it should be a lot easier to fix that, although I'm also fine fixing it as a follow-up to keep this change more behavior-preserving.

eventualbuddha: Good point. I added that too since, as you said, it wasn't too hard now that it's all async/await.

PR #1262, opened 2018-03-12
Add --bind-methods-after-super-call flag for a correct method binding.

anton-locomote:

### DIFF
--- a/script/update-website.ts
+++ b/script/update-website.ts
@@ -4,7 +4,7 @@ import { join } from 'path';
 import getLatestVersion from './getLatestVersion';
 
 let commit = true;
-let force = false;
+// let force = false;
 
 for (let i = 2; i < process.argv.length; i++) {
   switch (process.argv[i]) {
@@ -14,7 +14,7 @@ for (let i = 2; i < process.argv.length; i++) {
 
     case '--force':
     case '-f':
-      force = true;
+      // force = true;
       break;
 
     default:
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Here are some popular open source CoffeeScript projects and their current status
 when run through decaffeinate. Each project has a decaffeinate-specific fork
 that is re-created from the original repo once per day.
 
-Project                                                  | Lines of CoffeeScript | Conversion status                 | Test status                 
+Project                                                  | Lines of CoffeeScript | Conversion status                 | Test status
 -------------------------------------------------------- |:---------------------:|:---------------------------------:|:---------------------------:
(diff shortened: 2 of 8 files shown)


alangpierce:
Thanks for the PR. Would be good to discuss a bit. (Often times it makes most sense to file an issue to start a discussion before getting too far into the code, although this code change is small enough that a discussion in the PR is reasonable.)

This is actually how decaffeinate originally behaved, but it's incorrect, which is why I changed it to the current system. Here's an example that works in CoffeeScript but breaks if you put method bindings after the super call:
```coffee
listener = null

class B
  constructor: ->
    @initialize()

class A extends B
  f: => console.log('hello')

  initialize: ->
    listener = @f

new A()
listener()
```

In CoffeeScript 1 (which is what decaffeinate focuses on), this prints hello: [link](http://coffeescript.org/v1/#try:listener%20%3D%20null%0A%0Aclass%20B%0A%20%20constructor%3A%20-%3E%0A%20%20%20%20%40initialize()%0A%0Aclass%20A%20extends%20B%0A%20%20f%3A%20%3D%3E%20console.log('hello')%0A%0A%20%20initialize%3A%20-%3E%0A%20%20%20%20listener%20%3D%20%40f%0A%0Anew%20A()%0Alistener())

This actually came up in the codebase I was converting; running decaffeinate introduced a number of subtle bugs where handlers weren't being registered correctly.

In CoffeeScript 2, it crashes because the CS2 behavior puts it after super. CS2 support in decaffeinate is still a work-in-progress, and I think we'll need something like this change in order to properly follow CS2's behavior, but it would probably be behind the `--use-cs2` flag (which I added yesterday).

Some questions/thoughts:
* In the code that you're converting, are you using CS1 or CS2?
* What's your motivation for this change? Is the current behavior particularly painful for you?
* I think it might make sense to start the flag name with `--loose` since it's sacrificing correctness for ease of use.

script/update-website.ts
```diff
@@ -14,7 +14,7 @@ for (let i = 2; i < process.argv.length; i++) {
 
     case '--force':
     case '-f':
-      force = true;
+      // force = true;
```
eventualbuddha: Unintended changes?

hilts-vaughan:
This will also break a bunch of stuff with `Backbone`, for example since `Backbone`'s `initialize` call will potentially try and use "bound" methods which would not be bound yet, since `initialize` gets called during the `super` constructor... before binding can execute.

eventualbuddha:
Closing as stale. If there's a desire to discuss more we can reopen.

PR #1270, opened 2018-03-17
chore: run update-website as part of the release step

alangpierce:
Looks like this got dropped when moving to build stages.


alangpierce:
I was adding a CS2 option to the repl and noticed that it's still on 4.1.1, so looks like website publish hasn't been working. Hopefully this should fix.

Issue #1303, opened 2018-04-11
Babel / TS constructor hack gets optimized away using UglifyJS


hilts-vaughan:
Hi all!

If you put a random Coffeescript class through Decaf with Babel constructor hacks enabled, then run it through, say, here: https://skalman.github.io/UglifyJS-online/ with the constructor class. Say, with the following input:

```
class ConstructorHackDemo extends Backbone.Model {
  constructor(...args) {
    {
      // Hack: trick Babel/TypeScript into allowing this before super.
      if (false) { super(); }
      let thisFn = new Function('this');
      let thisName = thisFn.slice(thisFn.indexOf('{') + 1, thisFn.indexOf(';')).trim();
      eval(`${thisName} = this;`);
    }
    super(...args);
  }
}
```

... UglifyJS will then optimize part of the line away to read, say:

```
            let thisFn = (() => {}).toString(), thisName = thisFn.slice(thisFn.indexOf("{") + 1, thisFn.indexOf(";")).trim();
```

I think this is to be expected with stock settings. It persists even if you decide to to toggle `dead_code` or  `unused` I'll play around with it.

It seems to play better if you force it to think it's used, i.e:

```
let thisFn = (() => { return this; }).toString();
```

... and updating the below lines. Is this known? Not strictly an issue with decaf and I'm working around it right now but I noticed one of your goals was a "great out of the box experience" for many projects. :)

Or some other variant such as `new Function`

Great project 👍

hilts-vaughan:
... or switching to say, "new Function("this")" would fix this.. optimizer seems to leave it alone.

alangpierce:
Hi @hilts-vaughan, thanks for the report and the thoughts. For reference, here's the full constructor hack code snippet:
```
    {
      // Hack: trick Babel/TypeScript into allowing this before super.
      if (false) { super(); }
      let thisFn = (() => { this; }).toString();
      let thisName = thisFn.slice(thisFn.indexOf('{') + 1, thisFn.indexOf(';')).trim();
      eval(`${thisName} = this;`);
    }
```
There are two pieces here:
* Babel and TypeScript do static analysis checks to make sure that `super` at least has a chance of being called before the first use of `this`, so `if (false) { super(); }` tricks them into allowing it.
[...]

PR #1304, opened 2018-04-12
fix: Changed the constructor hack to be more explicit for minifiers

hilts-vaughan:
This is a fix for: https://github.com/decaffeinate/decaffeinate/issues/1303

I need to test on UglifyJS still and going to do that soon but wanted to get the feedback rolling. I took your suggestion with the validate and adding a test that would exercise the behavior but wasn't sure if you wanted to validate the actual constructor function itself. 

I regexp'd all the other tests to match the new lines as well, since there were a few. 

Queries:

1. Is this what you were thinking?
2. Otherwise, if you were not, did you expect to call `setResult` and verify the result of `thisName` somewhere? If so, how to verify that it ends up in the sandbox? I tried it but ending up getting `thisName` is undefined.

### DIFF
--- a/src/utils/babelConstructorWorkaroundLines.ts
+++ b/src/utils/babelConstructorWorkaroundLines.ts
@@ -23,8 +23,8 @@ export default [
   '{',
   '  // Hack: trick Babel/TypeScript into allowing this before super.',
   '  if (false) { super(); }',
-  '  let thisFn = (() => { this }).toString();',
-  "  let thisName = thisFn.slice(thisFn.indexOf('{') + 1, thisFn.indexOf(';')).trim();",
+  '  let thisFn = (() => { return this; }).toString();',
+  "  let thisName = thisFn.slice(thisFn.indexOf('return') + 6 + 1, thisFn.indexOf(';')).trim();",
   '  eval(`${thisName} = this;`);',
   '}',
 ];
--- a/test/class_test.ts
+++ b/test/class_test.ts
@@ -361,8 +361,8 @@ describe('classes', () => {
             {
               // Hack: trick Babel/TypeScript into allowing this before super.
               if (false) { super(); }
-              let thisFn = (() => { this; }).toString();
-              let thisName = thisFn.slice(thisFn.indexOf('{') + 1, thisFn.indexOf(';')).trim();
+              let thisFn = (() => { return this; }).toString();
+              let thisName = thisFn.slice(thisFn.indexOf('return') + 6 + 1, thisFn.indexOf(';')).trim();
               eval(\`\${thisName} = this;\`);
             }
             this.a = 2;
@@ -383,8 +383,8 @@ describe('classes', () => {
             {
               // Hack: trick Babel/TypeScript into allowing this before super.
               if (false) { super(); }
-              let thisFn = (() => { this; }).toString();
-              let thisName = thisFn.slice(thisFn.indexOf('{') + 1, thisFn.indexOf(';')).trim();
+              let thisFn = (() => { return this; }).toString();
+              let thisName = thisFn.slice(thisFn.indexOf('return') + 6 + 1, thisFn.indexOf(';')).trim();
               eval(\`\${thisName} = this;\`);
             }
             this.a = 2;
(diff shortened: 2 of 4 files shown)


hilts-vaughan:
The uglified code for the test case ends up being:

```
"use strict";

function _possibleConstructorReturn(t, e) {
    if (!t) throw new ReferenceError("this hasn't been initialised - super() hasn't been called");
    return !e || "object" != typeof e && "function" != typeof e ? t : e;
}

function _inherits(t, e) {
    if ("function" != typeof e && null !== e) throw new TypeError("Super expression must either be null or a function, not " + typeof e);
    t.prototype = Object.create(e && e.prototype, {
        constructor: {
            value: t,
            enumerable: !1,
            writable: !0,
            configurable: !0
        }
    }), e && (Object.setPrototypeOf ? Object.setPrototypeOf(t, e) : t.__proto__ = e);
}

function _classCallCheck(t, e) {
    if (!(t instanceof e)) throw new TypeError("Cannot call a class as a function");
}

var A = function t() {
    _classCallCheck(this, t), null == this.value && (this.value = 1), this.value += 1;
}, B = function(_A) {
    function B() {
        var _this;
        _classCallCheck(this, B);
        var _this, thisFn = function() {
            return _this;
        }.toString(), thisName = thisFn.slice(thisFn.indexOf("return") + 6 + 1, thisFn.indexOf(";")).trim();
        return eval(thisName + " = this;"), _this.value = 3, _this = _possibleConstructorReturn(this, (B.__proto__ || Object.getPrototypeOf(B)).apply(this, arguments));
    }
    return _inherits(B, _A), B;
}(A), b = new B();

// Get the value of b
console.log(b.value);
```

... and if you do

```
node uglify.js
```

... where `uglify.js` is this source code, you get `4` just like the unit test, which is a pass. Too bad there's no way  to integrate that into the build but that's not really the job of this project, I suppose. :)

test/class_test.ts
```diff
@@ -1952,4 +1952,20 @@ describe('classes', () => {
...
+          @value ?= 1
+          @value += 1
+      
+      class B extends A
+        constructor: ->
+          @value = 3
+          super
+      b = new B()
+      setResult(b.value)
+    `, 4, {options: {disableBabelConstructorWorkaround: false}, skipNodeCheck: true});
```
alangpierce: `disableBabelConstructorWorkaround` is false by default, so it should be unnecessary here, and I think the test name makes it clear enough.

test/class_test.ts
```diff
@@ -1952,4 +1952,20 @@ describe('classes', () => {
       })();
     `);
   });
+
+  it('produce a proper hack to trick Babel and into TypeScript that would be OK with UglifyJS -- and continues to produce results as would be expected', () => {
+      validateCS1(`
```
alangpierce: Yep, this is the sort of test that I had in mind, thanks!

(This line is indented two much by two spaces, though, I think.)

test/class_test.ts
```diff
@@ -1952,4 +1952,20 @@ describe('classes', () => {
       })();
     `);
   });
+
+  it('produce a proper hack to trick Babel and into TypeScript that would be OK with UglifyJS -- and continues to produce results as would be expected', () => {
```
alangpierce: This test name is a bit long. There isn't a strict line length limit, but ~100 chars is typically a good target. Also ideally the test name completes a sentence starting with "it".

Maybe just something like "produces a Babel/TS constructor hack that allows this before super in a constructor". Really this is a general test for correctness for the hack stuff.

As for the uglify thing, I actually maybe wouldn't mention it here since this test doesn't actually check the uglify output. But it sounds like a good thing to mention in the comment at the top of babelConstructorWorkaroundLines.ts.

hilts-vaughan: Agreed, and the fact that minifiers treat it properly is just an implementation detail of this.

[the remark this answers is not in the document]
hilts-vaughan: Yep and going to remove it all together -- thanks for the proofread.

test/class_test.ts
hilts-vaughan: Yep, fixed.

src/utils/babelConstructorWorkaroundLines.ts
```diff
@@ -23,8 +23,8 @@ export default [
   '{',
   '  // Hack: trick Babel/TypeScript into allowing this before super.',
   '  if (false) { super(); }',
-  '  let thisFn = (() => { this }).toString();',
-  "  let thisName = thisFn.slice(thisFn.indexOf('{') + 1, thisFn.indexOf(';')).trim();",
+  '  let thisFn = (() => { return this; }).toString();',
```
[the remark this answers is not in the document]
hilts-vaughan: Hope you don't mind. :)

PR #1310, opened 2018-04-26
Fixes: Updated constructor hack to be more specific

hilts-vaughan:
Firefox 45 (LTS) produces the following code, something like this:

```
use strict;
return _this;
```

Since the `use strict`; is there, it throws the hack off. With this patch / fork, we have successfully deployed to QA servers and ran against every major browser (all the way down to IE7) You can technically ship a product with this hack in place on every major browser now + old ones. 

We're currently in the process of removing all the hacks but it's good to be able get something out of the box that you can iterate on + run all your tests again in all the different browsers as you slowly chip away.

### DIFF
--- a/src/utils/babelConstructorWorkaroundLines.ts
+++ b/src/utils/babelConstructorWorkaroundLines.ts
@@ -24,7 +24,7 @@ export default [
   '  // Hack: trick Babel/TypeScript into allowing this before super.',
   '  if (false) { super(); }',
   '  let thisFn = (() => { return this; }).toString();',
-  "  let thisName = thisFn.slice(thisFn.indexOf('return') + 6 + 1, thisFn.indexOf(';')).trim();",
+  "  let thisName = thisFn.slice(thisFn.indexOf('return') + 6 + 1, thisFn.lastIndexOf(';')).trim();",
   '  eval(`${thisName} = this;`);',
   '}'
 ];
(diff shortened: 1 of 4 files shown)


src/utils/babelConstructorWorkaroundLines.ts
```diff
@@ -20,11 +20,11 @@
...
  */
 export default [
-  '{',
-  '  // Hack: trick Babel/TypeScript into allowing this before super.',
-  '  if (false) { super(); }',
-  '  let thisFn = (() => { return this; }).toString();',
-  "  let thisName = thisFn.slice(thisFn.indexOf('return') + 6 + 1, thisFn.indexOf(';')).trim();",
-  '  eval(`${thisName} = this;`);',
-  '}',
+  "{",
```
hilts-vaughan: Sorry, I think prettier did this... will ammend shortly to use single quotes.

hilts-vaughan:
Gah. Let me know if the prettier output bothers you can and I can disable it for the project. I'm trying it out newly and it's pretty annoying for projects that DON'T use it.

alangpierce:
@hilts-vaughan I just updated the code to use prettier. Would you mind rebasing?

hilts-vaughan:
Sure, I can do that. Give me a few.

hilts-vaughan:
Should be good. FYI, we shipped code like this for IE7, IE8, IE9, IE10, IE11, Edge, Firefox 45+, Chrome, Android and iOS. We're about half-way removing all the cruft and doing code reviews for the rest.

Thanks for all your help!

alangpierce:
@hilts-vaughan looks like cli_test.ts isn't quite prettified, see the error message in the test failure. Could you fix?

hilts-vaughan:
Yes, I'll run the tests locally this time too.

PR #1317, opened 2018-05-25
add output flag

idleberg:
Allows to set an output directory for the transpiled files. The `normalize` is required for nested folders that contain spaces.

I didn't know where to put the `--output` inn the help description, maybe the entire help output should be alphabetized anyway.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1,7 +1,7 @@
 /* eslint-disable no-process-exit */
 
-import { readdir, readFile, stat, writeFile } from 'mz/fs';
-import { basename, dirname, extname, join } from 'path';
+import { exists, mkdir, readdir, readFile, stat, writeFile } from 'mz/fs';
+import { basename, dirname, extname, join, normalize } from 'path';
 import { convert, modernizeJS } from './index';
 import { Options } from './options';
 import PatchError from './utils/PatchError';
@@ -24,12 +24,14 @@ interface CLIOptions {
   paths: Array<string>;
   baseOptions: Options;
   modernizeJS: boolean;
+  output: string;
 }
 
 function parseArguments(args: Array<string>): CLIOptions {
   let paths = [];
   let baseOptions: Options = {};
   let modernizeJS = false;
+  let output = '';
 
   for (let i = 0; i < args.length; i++) {
     let arg = args[i];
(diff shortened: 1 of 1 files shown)


src/cli.ts
```diff
@@ -24,12 +24,14 @@ interface CLIOptions {
...
   baseOptions: Options;
   modernizeJS: boolean;
+  output: string;
 }
 
 function parseArguments(args: Array<string>): CLIOptions {
   let paths = [];
   let baseOptions: Options = {};
   let modernizeJS = false;
+  let output = '';
```
eventualbuddha: I'd prefer to keep this as `null` or `undefined` when it has no explicit value, which means something like `let output: string | undefined;`.

src/cli.ts
```diff
@@ -181,7 +188,10 @@ async function runWithPaths(paths: Array<string>, options: CLIOptions): Promise<
 
   async function processFile(path: string): Promise<void> {
     let extension = path.endsWith('.coffee.md') ? '.coffee.md' : extname(path);
-    let outputPath = join(dirname(path), basename(path, extension)) + '.js';
+    if (options.output && !(await exists(options.output))) {
+      await mkdir(normalize(options.output));
```
eventualbuddha: This will only create the root directory, but nothing else. This will work well enough if converting a single file, but for converting a directory structure the `writeFile` call in `processFile` will fail for any files in a nested directory.

It would make this much easier to add to `cli_test.ts` to make sure that the cases you care about work as expected.

src/cli.ts
```diff
@@ -24,12 +24,14 @@ interface CLIOptions {
   paths: Array<string>;
   baseOptions: Options;
   modernizeJS: boolean;
+  output: string;
```
eventualbuddha: To follow the advice I gave about making `output` be potentially `undefined`, this should be `output?: string;`.

src/cli.ts
eventualbuddha: I’m not sure where that specific error would come from, but it looks like you’re trying to assign a boolean to `output`.

idleberg:
Sorry for the slow response. Should the tests be added to `cli_test.ts`?

alangpierce:
@idleberg yep! I'd just add a new test to the bottom of that file, and base it off of other tests. You'll probably want to make use of the test_fixtures directory, and feel free to add anything useful there for your tests.

eventualbuddha:
Closing as stale. If anyone wants to jump back in we can reopen.

PR #1333, opened 2018-10-11
fix Babel/TypeScript `super` hack to work w/ istambul

ryankshaw:
we use istambul in our tests to calculate code coverage for our app.
when it runs it wraps each statement with some stuff to track what lines
were executed. when it does that it adds an extra semicolon to:
`let thisFn = (() => { return this; })`
so this change it to have it just look for the last ";"
(the one right after `this`)


eventualbuddha:
I believe this is superseded by https://github.com/decaffeinate/decaffeinate/pull/1340 and is therefore fixed in v5.1.0 and later.

PR #1340, opened 2018-11-22
feat: add support for babel v7

eventualbuddha:
In babel v7 there is an additional check that `this` has been initialized in a constructor that calls `super`, which changes what we need to look for in the `toString`'d function we emit when using the Babel/TS `this` before `super` workaround.


eventualbuddha:
Thanks, @alangpierce 😄

PR #1445, opened 2019-07-24
types: remove unnecessary `as any` cast

eventualbuddha:
The typings for `@babel/parser` have been updated to account for options.


eventualbuddha:
Need to update the version of `@babel/parser` first to (maybe?) `7.5.5`.

eventualbuddha:
Outdated.

PR #2403, opened 2021-07-21
feat: support ES2021 logical assignment operators

shaylew:
In the spirit of `--optional-chaining`, this adds a new command-line option `--logical-assignment` that transpiles CoffeeScript's `&&=`, `||=`, and `?=` to ES2021's `&&=`, `||=`, and `??=`.

(Note that I couldn't get tests to pass on `master` without some tweaks. I will drop the initial unrelated commits and rebase onto master when that's fixed, but for now I thought it was more useful to have the branch actually build successfully so it could be tested.)


eventualbuddha:
This is great! I don't have time to review now (vacation), but will when I get back.

qnighy:
I would appreciate it if it is reviewed & merged...

Issue #2141, opened 2020-11-07
fat arrow functions bound before calls to super render constructors invalid


smurp:
decaffeinate is producing the wrong JavaScript based on my CoffeeScript input:

```coffee
class SuperClass
  constructor: () ->
    @a = 1
class SubClass extends SuperClass
  fatArrowFunction: () =>
    @b = 1
    return
```
I get this output:

```js
/*
 * decaffeinate suggestions:
 * DS002: Fix invalid constructor
 * Full docs: https://github.com/decaffeinate/decaffeinate/blob/master/docs/suggestions.md
 */
class SuperClass {
  constructor() {
    this.a = 1;
  }
}
class SubClass extends SuperClass {
  constructor(...args) {
    this.fatArrowFunction = this.fatArrowFunction.bind(this);
    super(...args);
  }

  fatArrowFunction() {
    this.b = 1;
  }
}
```

Here's what I expect it to be instead:

```js
class SuperClass {
  constructor() {
    this.a = 1;
  }
}
class SubClass extends SuperClass {
  constructor(...args) {
    super(...args);
    this.fatArrowFunction = this.fatArrowFunction.bind(this);
  }

  fatArrowFunction() {
    this.b = 1;
  }
}
```

eventualbuddha:
Previously decaffeinate would produce code that worked around this when using Babel or TypeScript to transpile to ES5 or lower, but Babel changed in a way that broke the workaround. Given that the workaround didn't work when using real ES6 classes, I just decided to remove it in https://github.com/decaffeinate/decaffeinate/commit/14ab7975138cc63366324a07956a7a30f78903c3.

Note that the output you posted does indicate that you should fix the invalid constructor. Most of the time you can simply change it manually to be as you wrote it, but it's possible that you'd been relying on the ordering that bound the method first, so that's how it leaves it.

eventualbuddha:
That's a fair point, though the only way I can think of to fix this is to use ES5-style classes. That's almost certainly more work than anyone is willing to put in.

smurp:
I know nothing about decaffeinate's internals -- forgive me -- but given that decaffeinate introduces the `.bind(this)` lines itself, can't it have control over whether they come before the call to `super`?

eventualbuddha:
[...]

PR #2420, opened 2022-04-20
fix: place this assignments after super under CS2

qnighy:
CoffeeScript 2 puts autogenerated `this` assignments after the call to the `super`-constructor.

For example, in this code:

```coffee
class A extends B
  constructor: (@a) ->
    super()
  f: =>
```

The assignments to `this.a` and `this.f` are generated just after `super()`.

This commit ensures that decaffeinate follows the same rule.


eventualbuddha:
Sorry, @qnighy, I missed this when I was looking at #2141. Is there anything here that didn't get fixed by #2421?

qnighy:
In the example above, `this.f` was fixed by #2141 and `this.a` not.

PR #2481, opened 2022-07-10
feat: add `bare` option to allow IIFE wrapper

### LINKED ISSUE #2412: Option to wrap decaffeinate output in IIFE
I have a feature request.

There should be an option `--iife` to wrap decaffeinate's code output (below the decaffeinate suggestions comment, if any) in an IIFE. The CoffeeScript compiler wraps its JS output in an IIFE, and decaffeinate should too, or at least there should be an option to do so automatically.

eventualbuddha:
Adds `bare` option, which is `true` by default, but when `false` will wrap the output in an IIFE similar to the official CoffeeScript compiler's default behavior. This can be triggered on the CLI with `--no-bare`, which mirrors `coffee`'s `--bare` flag.

**Example:**
```sh
❯ echo 'a = 1' | decaffeinate --no-bare
(function() {
const a = 1;

}).call(this);                                                                                                           
```

Closes #2412

### DIFF
--- a/test/cli_test.ts
+++ b/test/cli_test.ts
@@ -30,7 +30,8 @@ async function runCli(
   args: ReadonlyArray<string>,
   stdin: string,
   expectedStdout: string,
-  expectedStderr = ''
+  expectedStderr = '',
+  expectedExitCode = 0
 ): Promise<void> {
   if (stdin[0] === '\n') {
     stdin = stripSharedIndent(stdin);
@@ -67,7 +68,7 @@ async function runCli(
 
   // Check the exit code and output streams.
   expect({ exitCode, stdout: stdout.trim(), stderr: stderr.trim() }).toEqual({
-    exitCode: 0,
+    exitCode: expectedExitCode,
     stdout: expectedStdout.trim(),
     stderr: expectedStderr.trim(),
   });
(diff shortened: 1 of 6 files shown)


test/cli_test.ts
```diff
@@ -561,4 +562,32 @@ describe('decaffeinate CLI', () => {
...
+    await runCli(['--no-bare'], '', `(function() {\n\n}).call(this);`);
+  });
+
+  it('cannot use --modernize-js with --no-bare', async () => {
+    await runCli(
+      ['--no-bare', '--modernize-js'],
+      '',
+      '',
+      `
+      cannot use --modernize-js with --no-bare
```
svicalifornia, line 576: Just curious why `--modernize-js` cannot be used with `--no-bare`?

svicalifornia (commented):
Thanks for this change! Just have one question about `--modernize-js` and `--no-bare`…

test/cli_test.ts
eventualbuddha: I guess it's fine. I misremembered that `--modernize-js` implied `--use-js-modules`, but it doesn't. https://github.com/decaffeinate/decaffeinate/pull/2482

PR #2514, opened 2022-09-11
ci: only run `pull_request` trigger on forks

eventualbuddha:
From https://github.com/zopefoundation/meta/issues/145#issuecomment-1127295753


eventualbuddha:
This doesn't seem to be working.

PR #2679, opened 2023-09-04
Fix incorrect link in docs

drauf:
Link to cleanup suggestions was actually linking to the `How decaffeinate approaches correctness` page


eventualbuddha:
Thanks!

PR #2685, opened 2026-03-31
Add source map support

idleberg:
I understand `decaffeinate` to be a project primarily for converting codebases from CoffeeScript to JavaScript. However, I think that doesn't entirely rule out using `decaffeinate` for a transitional period, where one might want to be able to track runtime features to the CoffeeScript source. This is not possible without sourcemap support. This PR tries to implement this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "@babel/types": "^7.12.6",
     "@codemod/core": "^2.0.1",
     "@codemod/parser": "^1.2.1",
+    "@jridgewell/remapping": "^2.3.5",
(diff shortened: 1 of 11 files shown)


eventualbuddha:
Why not just use the official CoffeeScript compiler? I’m sure it has sourcemap support. I don’t think I understand the scenario where you’d use decaffeinate this way.

idleberg:
Decaffeinate uses the CoffeeScript compiler (or a fork of it) under the hood. Why is there no way to pass options all the way down to the compiler and leverage its features? I understand that my use-case for sourcempas might be considered a niche, but I want to look at this from a wider perspective.

